PR #17, opened 2017-08-04
Identifiers and locators - 1st draft (partial)

tcole3:
Adds definitions to 1.3 Terminology and text to 3.2.1 identifier metadata. More to come.

### DIFF
--- a/index.html
+++ b/index.html
@@ -145,6 +145,15 @@
 						<p>A user might follow alternative pathways through the content, but in the absence of such interaction
 							the default reading order defines the expected progression from one primary resource to the
 							next.</p></dd>
+
+					<dt><dfn data-lt="IRI">IRI</dfn></dt>
+					<dd><p>An IRI, or Internationalized Resource Identifier, is an extension to the URI specification to allow 
+						characters from Unicode. There is a mapping algorithm for translating between IRIs and the equivalent 
+						encoded URI form. IRIs are defined by [[!rfc3987]].</p></dd>
+						
+					<dt><dfn data-lt="Locator">Locator</dfn></dt>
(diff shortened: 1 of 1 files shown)


tcole3:
Possible edits to consider:

Would it be appropriate to add to the definition of Web Publication Canonical Identifier that it has global scope (i.e., is a global identifier as described by the _Architecture of the World Wide Web, Volume One_ W3C Recommendation, https://www.w3.org/TR/webarch/#identification), or is this so obvious today as to be unnecessary? 

The proposed definition of the Web Publication Canonical Identifier, which runs a little long, embeds mention of an 'Identifying IRI' - should that be its own definition?

There is still a bit of redundancy that could be deleted if it doesn't impact readability significantly.

iherman:
To make the PR more readable, here is a preview:

https://rawgit.com/tcole3/wpub/8b3d570bb3c2106e2fc754119369bd4a434067f3/index.html

(Note: eventually, there will be an automatic link to these previews and the corresponding diff-s in PR-s. The tool exists already, but there are some bugs right now that the relevant people are looking at...)

BigBlueHat:
@tcole3 great work getting this started and into the spec! I'm OK with use pushing this forward as the starting text. There are still (obviously) many surrounding issues to the "what's and wherefores" of the use of these identifiers and locators--and I'm hopeful we can clarify those usage patterns first, and then (if needed) revisit this (and any intervening) definition text.

I'd much rather change the translation book than the language. 😉

Thanks again, @tcole3!

iherman:
Discussed on the [meeting of the 7th of August](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2017/2017-08-07-minutes), and was merged as a result of the decision there.

PR #31, opened 2017-08-15
Normalize URL references

rdeltour:
Replaces instances of _IRI_ and _URI_ to _URL_, and added normative references to the [URL standard](https://url.spec.whatwg.org/).

I don't know how to add an explanatory note to the normative reference (fetched from SpecRef), this is beyond my ReSpec skills… @iherman any idea?

### DIFF
--- a/index.html
+++ b/index.html
@@ -151,11 +151,6 @@
 							the default reading order defines the expected progression from one primary resource to the
 							next.</p></dd>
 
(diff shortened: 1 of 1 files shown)


iherman:
@rdeltour I am in the process of consolidating the various comments and decisions; a merge will create a mess. I will take this version down on my disk and combine what is there manually, and close this PR without any further action. I hope that is fine. We should then look at the result when all editing are done (and reviewed by @mattgarrish :-).

PR #28, opened 2017-08-14
update identifier, address, canonical definitions

tcole3:
Per discussion of issue #27

PR #33, opened 2017-08-15
Consolidation of the draft

iherman:
* Refer to the publishing and linking document for references (see [resolution](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2017/2017-08-07-minutes#resolution3))
* Usage, in preference, of URL-s (see [resolution](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2017/2017-08-14-minutes#resolution3)). This also include the incorporation (manually) of the changes in PR #31.
* Reconciling the terminology section, after taking on board PRs #28 and #17
* Incorporated the content of the [separate document](https://rawgit.com/w3c/wpub/manifest-consensus-proposal/index-manifest-proposal.html#abstract-versus-concrete-manifest) as discussed on the [meeting at 2017-08-14](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2017/2017-08-14-minutes) (see [resolution](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2017/2017-08-14-minutes#resolution2)). In doing so, some earlier explanations on identifiers have been removed and/or moved elsewhere, to avoid duplication.
* Refreshed the list of participants in the acknowledgement section.

### DIFF
--- a/index.html
+++ b/index.html
@@ -325,6 +325,47 @@
 				<p class="issue" data-number="32"></p>
 			</section>
 
+			<section id="fallback">
+				<h3>Fallback mechanisms</h3>
+
+				<p>In some cases, some manifest items, even if required for an <a>abstract manifest</a>, are not necessarily expressed in a <a>concrete manifest</a> explicitly. For some of these cases “fallback mechanisms” are defined, i.e., simple algorithms whereby a <a class="externalDFN" href="https://www.w3.org/TR/publishing-linking/#dfn-user-agent">user agent</a> can reconstruct the manifest item. These are:</p>
+
+				<dl>
+					<dt>title</dt>
+					<dd>If the title is not specified in the concrete manifest, a user agent MUST:
+						<ul>
+							<li>
+								if a WP contains a single primary resource with a <code>title</code> element appropriate for its type (e.g., SVG or HTML titles), then that becomes the WP’s title;
+							</li>
+							<li>
+								otherwise, if a WP contains several primary resources with title elements appropriate for their types, the user agent uses the title of the first resource in the <a>default reading order</a>;
+							</li>
+							<li>
+								otherwise, the UA uses its own heuristics.
+								Note that a URL or a filename is not considered to be a valid title.
+							</li>
+						</ul>
+					</dd>
+					<dt>(natural) language</dt>
+					<dd>If the language is not specified in the concrete manifest, a user agent MUST:
+						<ul>
+							<li>
+								if a WP contains a single primary resource with a language element appropriate for its type (e.g., SVG or HTML language tags), then that becomes the WP’s language;
+							</li>
+							<li>
+								otherwise, if a WP contains several primary resources with a language element appropriate for their types, the user agent uses the language of the first resource in the default reading order;
+							</li>
+							<li>
+								otherwise, the language is English.
+							</li>
+						</ul>
+					</dd>
+				</dl>
+			</section>
+
+
+
+
 			<section id="metadata">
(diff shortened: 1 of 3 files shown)


iherman:
This is just a draft consolidation. It will need lots of care from @mattgarrish !

iherman:
The text can be seen rendered through proper HTML (and respec) at:

https://rawgit.com/w3c/wpub/6f47335caf1c02fb97b4be100f2f5fd366bf3c70/index.html

index.html
```diff
@@ -143,54 +143,82 @@
...
-					<dt><dfn>Default Reading Order</dfn></dt>
-					<dd><p>The default reading order is a specific progression through the <a>primary resources</a> defined in
-						the <a>manifest</a> by the creator of a <a>Web Publication</a>.</p>
-						<p>A user might follow alternative pathways through the content, but in the absence of such interaction
-							the default reading order defines the expected progression from one primary resource to the
-							next.</p></dd>
+					<dt><dfn>URL</dfn></dt>
+					<dd>
+						<p>
+							In this specification, the general term URL is used as defined and used in other W3C specification like HTML&nbsp;[[html51]], and is defined by the relevant WhatWG document&nbsp;[[url]].
```
rdeltour: * editorial comments:
  * `s/is used as defined and used/is used as/`
  * `s/in other W3C specification/in other W3C specifications/`
  * `s/the relevant WhatWG document/the URL Standard from WhatWG/`
* the ref to the URL standard is normative, so it should be `[[!URL]]`, no?
* shouldn't we use an unversionned reference to the HTML spec?  (https://www.w3.org/TR/html/).

index.html
```diff
@@ -143,54 +143,82 @@
...
+							In particular, such a URL allows for the usage of characters from Unicode following&nbsp;[[rfc3987]].
+							See <a href="https://www.w3.org/TR/html51/references.html#biblio-url">the note in the HTML5 document</a> for further details.
```
rdeltour: same question about using an unversionned reference

index.html
iherman: Thanks. Comments on comments:
* I have changed to "Standard from WhatWG", we shall see what the W3C publishing rules will say about this.
* The reference to URL is in a non-normative section, hence the reference itself cannot be normative (per pubrules...)

Otherwise done.

index.html
iherman: Done.

iherman:
Regenerated a new preview, to include the changes for @rdeltour 

https://rawgit.com/w3c/wpub/816fee6c3c8ba834dd0e22adc11e380e69950cae/index.html

index.html
rdeltour: > The reference to URL is in a non-normative section, hence the reference itself cannot be normative (per pubrules...)

ah, of course :) thanks!

mattgarrish:
One general comment is to try and reduce the acronymizing of the document. WP, UA, TOC might make typing easier, but they don't enhance the readability.

I also don't see adding "Web Publication" before some of the terms adds anything, and only leads to trying to shorten (A Web Publication has an address, for example, or do we intend to write somewhere about the Web Publication's Web Publication Address?)

I would argue, in fact, that the "WP" definitions (Canonical Identifier, Identifier and Address) don't belong in the terminology at all, but should be defined where they are first used, especially since canonical identifier introduces normative requirements into terminology. We're not creating these terms, only explaining their relationship to a web publication.

But these nits can all be smoothed out later.

I don't have anything to add to the substance of the PR.

PR #37, opened 2017-08-17
Modification of the manifest abstract model (proposal)

llemeurfr:
It only conflates requirements 3 and 8 in 3, because the sequence of primary resources will defines the default reading order, whatever concrete syntax is chosen. It also adds a note about the Web Publication address, which references the requirements this property fulfills (I think that in the current editor's version, we should reference requirements and maybe create some new ones).

### DIFF
--- a/index.html
+++ b/index.html
@@ -275,21 +275,21 @@
 
 				<ol>
 					<li>It MUST assert WP-ness.</li>
-					<li>It MUST list the <a>Primary Resources</a> of the WP.</li>
-					<li>It SHOULD list <a>Secondary Resources</a>, although that list is not necessarily exhaustive.</li>
 					<li>
-						It MUST include a <a>WP Address</a>.
+						It MUST include a <a>Web Publication Address</a>.
 						Availability of this address does not preclude the creation and use of other identifiers and/or addresses to retrieve a representation of a <a>Web Publication</a> in whole or part.
-					</li>
+					</li>					
+					<li>It MUST list the <a>Primary Resources</a> of the WP in their default reading order.</li>
+					<li>It SHOULD list <a>Secondary Resources</a>, although that list is not necessarily exhaustive.</li>
 					<li>
 						It SHOULD include a <a>Canonical Identifier</a>.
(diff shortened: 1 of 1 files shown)


mattgarrish:
I agree with Tzviya that this seems like a conflation of abstract and concrete. A concrete could fulfill the requirements in any number of ways, as I understand the current proposal.

llemeurfr:
I would agree we you if we were listing requirements, which must be as independent as possible. But if this abstract model is what we often call a conceptual model, i.e. something which can be represented by an Entity Relationship Diagram (or UML class diagram to be more pragmatic), the use of ordered lists of entities as a way to specify ... ordered list of entities, seems simple and logical. But it's not a big deal if the consensus of the WG is not favorable for this minor modification, as the abstract model will certainly find a place outside of the final spec. 

I'd like to have feedback on the addition of notes associating the abstract model statements to the corresponding requirements (for the IG plus the lists we should maintain as a WG).

iherman:
@llemeurfr, in the abstract sense the primary resources do not necessarily form a list. It may just be a set as far as its goal is concerned. If we use JSON, then of course it is true that the only way to represent them is to use a list and not a set, but that is the specificity of JSON. If, God forbid!, we were to use RDF to represent that data, then it would be just about the opposite: while it is easy to represent a set of statements, it becomes difficult to represent a list. Ie, I think it is better to keep the two notions separate.

(Before somebody jumps at me, I do _not_ propose to use RDF for that purpose, it is just an example...)

iherman:
@llemeurfr the note you propose is fine but, personally, I would prefer to do this, at some point differently and more systematically. It would indeed be good to check, _and document_, how the various UCR requirements are covered by the entries in the FPWD and if they are not, why. I would propose that we would do that as, e.g., an informal appendix of the document. But I am a bit afraid it may be too early to do that... (I am a bit afraid that if we did that systematically in the document proper, it would make the text very difficult to read.)

That being said, we can indeed add the skeleton for such an appendix into the document and put your two examples there.

WDYT?

llemeurfr:
I'll close this PR for now, to clear the discussions.

PR #41, opened 2017-08-20
Run ReSpec sync instead

marcoscaceres:
Should avoid race conditions.... otherwise try:

```HTML
<script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove" defer></script>
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,7 +3,7 @@
 	<head>
 		<meta charset="utf-8" />
 		<title>Web Publications</title>
-		<script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove" async></script>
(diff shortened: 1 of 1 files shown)


iherman:
Marked as non substantive for IPR from ash-nazg.

iherman:
Actually, what did the trick was the usage of `defer`.

marcoscaceres:
Great to hear that one of the solutions worked. Frustrated that I wasn't able to reproduce the bug 😿

iherman:
> Great to hear that one of the solutions worked. Frustrated that I wasn't able to reproduce the bug 😿
> 
> 

Wouldn't life be boring otherwise?

marcoscaceres:
Indeed 🍾

Issue #39, opened 2017-08-18
Do all documents in the reading order have to be reachable from the ToC


rdeltour:
_edited to better describe the scope_

There is a consensus that a Web publication must have a _reading order_ (a list of primary resources) and must/should have a _table of contents (ToC)_ (the main navigation entry point).

## Scope

The goal of this issue is to debate whether accessibility best practices require that **all** the primary resources should be reachable from the ToC, or not. This is about conceptual relationship between the ToC and reading order.

In other words, the objective is to provide answers to the following questions (the list can be edited as the discussion goes on):

1. are there legitimate use cases where a primary resource would not be listed in the ToC
2. a primary resource is not listed in the ToC, does it violate [WCAG SC 2.4.5](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-mult-loc)
3. if not, and if the ToC is incomplete, is the list of primary resource equivalent of a "site map" (i.e. does it satisfy [WCAG technique G63](https://www.w3.org/TR/WCAG20-TECHS/G63)?
4. ultimately, how does that translate into the spec (must, should, note, a11y guidelines, etc.)

## Out of scope

This issue is **not** about discussing:
- the format details
- how to infer one structure from another (as in issue #36)

As [@baldurbjarnason pointed out](https://github.com/w3c/wpub/issues/36#issuecomment-323151953):

> we need to start off by specifying a format that supports these as independent structures before we specify how you'd use one type of structure to generate another. And we should be explicit in the specification that these conversions can fail for a variety of reasons.

## References

- Issue #38 "What does “Reading Order” mean"
- Issue #36 "Is the ToC sufficient to provide reading order?"
- Issue #26 "Should the manifest be an implicit ToC?"
- [WCAG SC 2.4.5 "Multiple ways"](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-mult-loc)
- [Understanding WCAG SC 2.4.5](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-mult-loc.html)

rdeltour:
In issue #36, @HadrienGardeur proposed [an example where the has 6 different primary resources but the fallback TOC could consist of only three entries](https://github.com/w3c/wpub/issues/36#issuecomment-323210747).

I pointed that this could be a violation of [WCAG SC 2.4.5 "Multiple ways"](https://www.w3.org/TR/WCAG20/#navigation-mechanisms-mult-loc).

[...]

Issue #36, opened 2017-08-17
Is the ToC sufficient to provide reading order?


BigBlueHat:
Relates to both #26 "Should the manifest be an implicit TOC?" and the more recent #35 "Proposal: an HTML-first Table of Contents approach to Web Publication."

HadrienGardeur:
I'll re-post what I already said in #35.

A TOC is not sufficient to provide a reading order because it's a completely different concept.

A TOC can:

- only point to some, not all primary resources (for example it might skip preliminary resources such as a copyright page, or a resource containing a single quote)
- point to the same resource multiple times (for example using fragments to specific locations in a resource)
- point to resources in an order that's not the reading order (for example a travel guide could contain alternate ways of browsing the publication based on location or a specific interest)

A reading order would:

- reference all primary resources, and reference them once
- strictly follow the reading order, and not provide other methods for navigating in a publication

baldurbjarnason:
My problem with this issue as well as with the proposal in issue https://github.com/w3c/wpub/issues/35 is that it sacrifices the specificity and clarity of the manifest as aformat to aid in the authorship of a very specific kind of publication.

The manifest, once it exists as a concrete structure in an application, _absolutely_ needs reading order, table of contents, and secondary resources as separate structures. If the format we specify does not specify these structures as separate pieces of data, we are going to have issues with both interoperability and with our support for diverse forms and genres of publications.

So the manifest format absolutely does need to specify these as separate structures in the format from the start. This is something that the html-first proposal does not do and is the root cause for why it is problematic in general (IMO).

Now, I do think that if the reading order is absent, the UA should use the ToC to generate it, and if the ToC is absent, the UA should use the reading order to generate that. And if a list of secondary resources is missing, a UA may want to try to generate that list through some other means. I think these are important features we should at least try to support to help authors do their job.

But we need to start off by specifying a format that supports these as independent structures before we specify how you'd use one type of structure to generate another. And we should be explicit in the specification that these conversions _can fail_ for a variety of reasons.

iherman:
First of of all, the very fact that this question comes to the fore means that, in the current list of items in the abstact manifest, the TOC is missing. It is conceptually different from the other items, and we should add it. @mattgarrish?

[...]

Issue #35, opened 2017-08-16
Proposal: an HTML-first Table of Contents approach to Web Publication


BigBlueHat:
@dauwhe and I have been working on a proposal to use HTML's `<nav>` element as a web publication manifest: https://github.com/dauwhe/html-first

TL:DR define the primary resources of a WP to be the files referenced in the first <nav> element of an "index" file. This file would also host WP metadata.

We feel this approach has many benefits:

1. Human-focused. User agents need a list of primary resources and their default ordering, but so do actual users. Most web publications would benefit from a human-readable table of contents. TOCs are crucial for accessibility. 

2. Simplicity. Given the broad need for a TOC, using that as manifest is a straightforward way to avoid duplication (as in EPUB's nav/manifest/spine/ncx). And we've discovered a huge benefit, as we don't need a list of secondary resources to facilitate offline caching via service workers (see the demo books)!

3. Ubiquity. Everyone in the web space is already familiar with HTML, and there is a large and mature ecosystem around authoring, rendering, and validating HTML.

4. Expressiveness. HTML's language and styling support allows for a richer experience for humans. 

5. Progressive enhancement. Existing web user agents know what to do with HTML. 

6. A Path to the future. Every EPUB3 has a nav document. Many "web books" already use such a design pattern. 

Note we've created a couple of [demo books](https://github.com/dauwhe/html-first#demo-books) that work offline, based on the HTML manifest. 

Thanks,
🎩 and 🐳

lrosenthol:
Very well thought out proposal - thanks for submitting.

Overall, I think it solves a lot of problems and I like the direction. I
also like the fact that an author that doesn't want/need it, can simply use
CSS to make it hidden and leave it strictly for the UA.

That said, I have a big concern with it - that it doesn't actually provide
a way to define "publication wide" information.  Let's consider the
discussion around Language definition (#29).  If I have a publication that
is multilingual, where some primary resources are in English and others are
in French (such as in parts of Canada) - I can't do that in your method
since the lang needs to be that of the resource in question.

If we can find a way to solve the publication-wide issue, then count me
in.  Otherwise, sorry...

BigBlueHat:
@lrosenthol great question! HTML saves the day again. 😄 
https://www.w3.org/TR/html52/dom.html#the-lang-and-xmllang-attributes

[...]

Issue #16, opened 2017-08-04
non-linear resources - primary, secondary or something else?


mattgarrish:
An issue I'm sure no one is looking forward to, but where do "non-linear" resources fall into the publication hierarchy?

For reference, non-linear is what epub calls documents that contain supplementary information not part of the primary narrative. In EPUB, they were marked as non-linear and required to be in the spine, but whether they were suppressed when reading the book or not was reading system dependent.

Is WP going to define something similar, or can the reading order be defined without including supplementary documents in the reading order?

If they can be excluded, these will be neither primary nor secondary resources, but something else.

If they can be excluded, does the author have to provide previous/next page navigation.

lrosenthol:
On Fri, Aug 4, 2017 at 9:05 AM, Matt Garrish <notifications@github.com>
wrote:

> An issue I'm sure no one is looking forward to, but where do "non-linear"
> resources fall into the publication hierarchy?
>
> For reference, non-linear is what epub calls documents that contain
> supplementary information not part of the primary narrative.
>
Can you give an example?   I've been trying to think of something
primary(!) that wouldn't go in the DRO, but can't.   Most of ones that I
can, would be secondary resources.


> If they can be excluded, does the author have to provide previous/next
> page navigation.
>
I think we decided that we were tabling the author-provided vs. UA provided
UX questions given their controversial nature...

mattgarrish:
> Can you give an example?

The general EPUB example is an answer key. You might provide a link to one where a quiz is, but you don't necessarily want it in the reading order. It's technically a primary resource, but would violate a statement that all primary resources are in the reading order.

Thinking of the web, any kind of document you would pop out rather than have as part of a sequential reading order has an equally ambiguous status. They're not secondary, but not primary, either. Do these have to be listed as primary resources in the reading order?

RachelComerford:
Agreed - we deal with this in education with glossary definitions that are
linked in the basal text in education (as well as footnotes, references,
etc)

Rachel Comerford | Director of Content Standards | T 212.576.9433

*Macmillan Learning*

lrosenthol:
On Fri, Aug 4, 2017 at 9:46 AM, Matt Garrish <notifications@github.com>
wrote:

> Can you give an example?
>
> The general EPUB example is an answer key. You might provide a link to one
> where a quiz is, but you don't necessarily want it in the reading order.
> It's technically a primary resource, but would violate a statement that all
> primary resources are in the reading order.
>
[...]

PR #46, opened 2017-08-23
default reading order fallbacks via TOC

iherman:
This is my attempt to extract from a number of issues (#35, #36, #39) some aspects of TOC vs. reading orders that may represent a level of consensus. I was mostly inspired by the comment of @baldurbjarnason (https://github.com/w3c/wpub/issues/36#issuecomment-323151953) which described that there _may_ be a fallback both to a TOC if a default reading order is not present, and on default reading orders if a TOC is not present, although both are listed as separate information in the WP Infoset.

Obviously, the trigger for all this was also #35

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/toc-html-fallback.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/9718f6a...fa23b0f.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -425,11 +425,11 @@
 				</p>
 
 				<p>
-					If the default reading order is not specified in the <a href="#manifest">manifest</a>, but the <a href="#table-of-contents">table of content</a> is available (either as part of the <a href="#manifest">manifest</a> or retrived from among the <a>primary resources</a>), the <a>primary resources</a> listed in the table of content also provide the default reading order (multiple references to the same <a>primary resource</a> in the table of content should be disregared in favour of the first occurence of that resource).
+					If the default reading order is not specified in the <a href="#manifest">manifest</a>, but the <a href="#table-of-contents">table of contents</a> is available (either as part of the <a href="#manifest">manifest</a> or retrived from among the <a>primary resources</a>), the <a>primary resources</a> listed in that table of contents also provide the default reading order (multiple references to the same <a>primary resource</a> in the table of contents should be disregared in favor of the first occurence of that resource).
 				</p>
 
 				<p class=ednote>
-					The relationship between the default reading order and the table of content is the subject of several issues; see the list in the <a href="#table-of-contents">section on table of contents</a>. 
+					The relationship between the default reading order and the table of contents is the subject of several issues; see the list in the <a href="#table-of-contents">section on table of contents</a>. 
 				</p>
 
 			</section>
@@ -437,25 +437,25 @@
 			<section id="table-of-contents">
 				<h3>Table of Contents</h3>
 
-				<p>If the table of content is not specified in the <a href="#manifest">manifest</a>, the user agent MUST make an attempt to retrive one as follows:</p>
+				<p>If the table of contents is not specified in the <a href="#manifest">manifest</a>, the user agent MUST make an attempt to retrive one as follows:</p>
 
 				<ol>
 					<li>
-						Check, among the <a>primary resources</a> of the publication, a resource with basename <code>index.html</code> if it exists, and try to extract a <a>TOC navigation element</a>;
+						Check, among the <a>primary resources</a> of the publication, an HTML resource with basename <code>index.html</code> if it exists, and extract a <a>TOC navigation element</a> from the resource if present;
 					</li>
 					<li>
-						otherwise, locate the first HTML resource among the <a>primary resources</a> in the order provided by the <a>default reading order</a> that includes a <a>TOC navigation element</a>.
+						otherwise, locate the first HTML resource among the <a>primary resources</a> in the order provided by the <a>default reading order</a> that also includes a <a>TOC navigation element</a>, and extract it.
 					</li>
 				</ol>
 
-				<p>If such <a>TOC navigation element</a> is found, its content provides the table of content information for the Web Publication infoset. If several of those are found, only the first one is considered.</p>
+				<p>If such <a>TOC navigation element</a> is found, its content provides the table of contents information for the Web Publication infoset. If several of those are found, only the first one is considered.</p>
 
 				<p>
 					A <dfn>TOC navigation element</dfn> is an HTML <code>nav</code> element with the <code>role</code> attribute set to <code>doc-toc</code>&nbsp;[[!dpub-aria-1.0]]</p>.
 				</p>
 
 				<p class=note>
-					This process may not yield a table of content, which is in line with the fact that a table of content is not a required information in the <a href="#infoset-req">WP infoset</a>.
+					This process may not yield a table of contents, which is in line with the fact that a table of contents is not a required information in the <a href="#infoset-req">WP infoset</a>.
 				</p>
 
 				<p class="issue">
@@ -463,7 +463,7 @@
 				</p>
 
 				<p class=ednote>
-					The issue of using the HTML <code>nav</code> element as a possible encoding of the table of content is mentioned or explicitly addressed in a number of issues listed below.
+					The issue of using the HTML <code>nav</code> element as a possible encoding of the table of contents is mentioned or explicitly addressed in a number of issues listed below.
 				</p>
 
 				<p class="issue" data-number=26></p>
--- a/index.html
+++ b/index.html
@@ -431,45 +431,17 @@
 				<p class=ednote>
 					The relationship between the default reading order and the table of contents is the subject of several issues; see the list in the <a href="#table-of-contents">section on table of contents</a>. 
 				</p>
+				<p class="issue" data-number=26></p>
+				<p class="issue" data-number=35>Define the primary resources of a WP to be the files referenced in the first</p>
+				<p class="issue" data-number=36></p>
+				<p class="issue" data-number=39>There is a consensus that a Web publication must have a reading order (a list of primary resources) and must/should have a table of contents (ToC) (the main navigation entry point).</p>
 
 			</section>
 
 			<section id="table-of-contents">
 				<h3>Table of Contents</h3>
 
-				<p>If the table of contents is not specified in the <a href="#manifest">manifest</a>, the user agent MUST make an attempt to retrive one as follows:</p>
-
-				<ol>
-					<li>
-						Check, among the <a>primary resources</a> of the publication, an HTML resource with basename <code>index.html</code> if it exists, and extract a <a>TOC navigation element</a> from the resource if present;
-					</li>
-					<li>
-						otherwise, locate the first HTML resource among the <a>primary resources</a> in the order provided by the <a>default reading order</a> that also includes a <a>TOC navigation element</a>, and extract it.
-					</li>
-				</ol>
-
-				<p>If such <a>TOC navigation element</a> is found, its content provides the table of contents information for the Web Publication infoset. If several of those are found, only the first one is considered.</p>
-
-				<p>
-					A <dfn>TOC navigation element</dfn> is an HTML <code>nav</code> element with the <code>role</code> attribute set to <code>doc-toc</code>&nbsp;[[!dpub-aria-1.0]]</p>.
-				</p>
-
-				<p class=note>
-					This process may not yield a table of contents, which is in line with the fact that a table of contents is not a required information in the <a href="#infoset-req">WP infoset</a>.
-				</p>
-
-				<p class="issue">
-					This question arises only if this mechanism is accepted: the question is whether a <a>TOC navigation element</a> can refer, via links, to any resource that is <em>not</em> listed as a <a>primary resource</a>.
-				</p>
-
-				<p class=ednote>
-					The issue of using the HTML <code>nav</code> element as a possible encoding of the table of contents is mentioned or explicitly addressed in a number of issues listed below.
-				</p>
-
-				<p class="issue" data-number=26></p>
-				<p class="issue" data-number=35>Define the primary resources of a WP to be the files referenced in the first</p>
-				<p class="issue" data-number=36></p>
-				<p class="issue" data-number=39>There is a consensus that a Web publication must have a reading order (a list of primary resources) and must/should have a table of contents (ToC) (the main navigation entry point).</p>
+				<div class="ednote">Placeholder for identifying table of contents - whether embedded or by reference.</div>
 				
 
 			</section>
--- a/index.html
+++ b/index.html
@@ -420,13 +420,58 @@
 			<section id="default-reading-order">
 				<h3>Default Reading Order</h3>
 
-				<p>The default reading order MUST include at least one <a>primary resource</a>.</p>
+				<p>
+					The default reading order MUST include at least one <a>primary resource</a>.
+				</p>
+
+				<p>
+					If the default reading order is not specified in the <a href="#manifest">manifest</a>, but the <a href="#table-of-contents">table of content</a> is available (either as part of the <a href="#manifest">manifest</a> or retrived from among the <a>primary resources</a>), the <a>primary resources</a> listed in the table of content also provide the default reading order (multiple references to the same <a>primary resource</a> in the table of content should be disregared in favour of the first occurence of that resource).
+				</p>
+
+				<p class=ednote>
+					The relationship between the default reading order and the table of content is the subject of several issues; see the list in the <a href="#table-of-contents">section on table of contents</a>. 
+				</p>
+
 			</section>
 
 			<section id="table-of-contents">
 				<h3>Table of Contents</h3>
 
-				<div class="ednote">Placeholder for identifying table of contents - whether embedded or by reference.</div>
+				<p>If the table of content is not specified in the <a href="#manifest">manifest</a>, the user agent MUST make an attempt to retrive one as follows:</p>
+
+				<ol>
+					<li>
+						Check, among the <a>primary resources</a> of the publication, a resource with basename <code>index.html</code> if it exists, and try to extract a <a>TOC navigation element</a>;
+					</li>
+					<li>
+						otherwise, locate the first HTML resource among the <a>primary resources</a> in the order provided by the <a>default reading order</a> that includes a <a>TOC navigation element</a>.
+					</li>
+				</ol>
+
+				<p>If such <a>TOC navigation element</a> is found, its content provides the table of content information for the Web Publication infoset. If several of those are found, only the first one is considered.</p>
+
+				<p>
+					A <dfn>TOC navigation element</dfn> is an HTML <code>nav</code> element with the <code>role</code> attribute set to <code>doc-toc</code>&nbsp;[[!dpub-aria-1.0]]</p>.
+				</p>
+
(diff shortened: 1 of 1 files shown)


HadrienGardeur:
I'm against the idea of an HTML TOC being a fallback for the primary reading order. 

These are two completely separate concepts, and while I feel that the primary reading order may be a fallback for a TOC, it's not true the other way around without additional requirements from the TOC.

iherman:
Adding additional requirement on a TOC for a fallback is perfectly fine, @HadrienGardeur. Let us formulate those and getting them documented. The goal remain to allow for very simple WP-s to be create-able easily, while letting the manifest control the complex cases.

HadrienGardeur:
@iherman but these additional requirements on a TOC only make sense when the TOC is a fallback. I'm not a big fan of that, it's better to have must/should/may for the TOC that are unrelated to its fallback status.

mattgarrish:
And these fallbacks seem to get more brittle as we try to construct ever more complex requirements. It's garbage in, garbage out processing.

I'm beginning to think these don't belong with the infoset but are guidance purely for a user agent to construct for an invalid publication. Legitimizing this for lazy authoring is a bad idea.

Anyway, some offered changes and thoughts.

As Hadrien has pointed out before, I believe, the same primary resource can occur multiple times legitimately, so you only want to ignore consecutive references to the same resource.

> If the default reading order is not specified in the manifest, the user agent MUST construct it from the table of contents, if a table of contents is provided. The default reading order becomes the list of primary resources linked to from the table of contents, in the order they occur, and after removing any fragment identifiers and de-duplicating any consecutive resources that remain.

The table of contents generation sounds like a bit of a fishing expedition in the hopes that it is there but the user forgot to link to it. It seems unlikely the UA will find a nice doc-toc marked nav element, though. (If only the first one found is used, then you can skip saying to disregard other ones.)

In terms of a fallback that is reasonably assured to be there, though, I thought the current thinking was to take the titles of the primary resources in the reading order? You'll at least get some kind of table of contents for purely image-based works, since you can fall back to the url or file name.

mattgarrish:
And just to be clear, what I mean by lazy authoring is not that there may be legitimate cases where this information could be reliably constructed. Rather, once we start assuming things will fall nicely into place because of certain scenarios we can envisage we end up with a very permissive specification that enables bad experiences for users for authors not aware of why the affordances are there.

We should have very tight rules and call this implicit information gathering (e.g., you don't have to list the default reading order, but you must set a "fromToc" flag) and/or push the processing into failure handling.

GarthConboy:
It does sound like there is general agreement that the HTML-supplied TOC could be a fallback for the case where the manifest doesn't include a specific TOC.   So, maybe this PR should be split in half, getting that part in?

I waffle on HTML TOC as fallback for manifest-resident primary reading order.  Two competing arguments:

-- Plus: makes authoring simple WPs easy, perhaps obviating the need for a JSON manifest for very simple WPs.
-- Minus:  The two are pretty different concepts, and "perhaps obviating the need for a JSON manifest for very simple WPs" is a bug that could lead to two valid types of "manifests".

This morning I was more on this plus side, now I'm more on the minus side.

HadrienGardeur:
I think @mattgarrish is making a good point that some of these fallbacks should be candidates for failure handling rather than the normal way we process a manifest.

Right now we have a very reasonable list of requirements for our infoset and I'm not comfortable with fallbacks that can barely handle them, especially for the primary reading order.

The lack of a list of primary resources in a manifest also raises the following question: if I discover the first chapter of a publication, and through it the manifest, how do I locate the TOC if the manifest does not contain any primary resource?

Do I need to rely on well-known location or the presence of a specific link in the content to figure out where the TOC lives?

This doesn't sound like "easy authoring" to me, and I'm not even talking about all the issues around processing a TOC to extract a reading order from it.

iherman:
Admin: I agree with the https://github.com/w3c/wpub/pull/46#issuecomment-324432427 of @GarthConboy : it was a mistake to lump two issues into one PR. So I split it:

1. This PR is on the section on default reading order, that is generated, as fallback, from a TOC
2. PR #47 is on the TOC being found in an HTML resource if not explicit in the manifest

iherman:
@GarthConboy summarized it as:

> I waffle on HTML TOC as fallback for manifest-resident primary reading order. Two competing arguments:
>
> -- Plus: makes authoring simple WPs easy, perhaps obviating the need for a JSON manifest for very simple WPs.
> -- Minus: The two are pretty different concepts, and "perhaps obviating the need for a JSON manifest for very simple WPs" is a bug that could lead to two valid types of "manifests".

I would not dismiss the second point. Instead of being very general, I take a document that, in my view, must be considered as a candidate for a WP, namely the HTML51 spec[1]. Although there is a single file version of the spec, the basic one is cut into a moderately large number of HTML files (35, to be precise). It also has a fairly detailed and long TOC which is, actually, repeated in all HTML files. The fact that it is repeated over all HTML files is a matter of the current Rec styling; a previous version of HTML[2] had its full TOC in a single file, namely Overview.html[3].

My goal is: I would like to avoid that producer of the HTML spec would have to maintain a _separate_ reading order in a manifest file. Or are we saying that, no matter what, authors _must_ produce this? I sense significant push-back coming from the editors of a document like HTML.

Looking at the TOC right now[4] the rough algorithm described in the draft works fairly well: extract the URL-s in order from the TOC, and remove duplicates. What you get is the default reading order…


1. https://www.w3.org/TR/html51/
2. https://www.w3.org/TR/2014/REC-html5-20141028/
3. https://www.w3.org/TR/2014/REC-html5-20141028/Overview.html#contents
4. https://www.w3.org/TR/html/Overview.html

HadrienGardeur:
@iherman this is only one example, not all table of contents look the same.

Since we haven't defined at all what a TOC must/should/may contain, this feels extremely premature  to have a TOC as an implicit fallback.

iherman:
> 
> @iherman <https://github.com/iherman> this is only one example, not all table of contents look the same.
> 
> Since we haven't defined at all what a TOC must/should/may contain, this feels extremely premature to have a TOC as an implicit fallback.
> 
> 


If it is a matter of being premature: I am fine with that insofar as we can have a separate discussion on the structure of the TOC. I think the FPWD should, nevertheless, make clear that this is indeed an issue that will be addressed at some point (whether it is with the wording of the current PR or not is a detail). I still believe we should make efforts to make it easy for documents like the HTML spec to be a bona fide WP with very low extra costs…

HadrienGardeur:
@iherman 

It's not the only issue that I have with it, but the fact that we haven't even defined:

- what a TOC truly is
- the requirements for a TOC
- how we identify a TOC

... is enough to at least postpone this discussion.

If a manifest doesn't contain a list of primary resources or a title, this means that we could have a WP without a manifest, which brings additional issues (How do we establish WP-ness without a manifest? How do we locate a TOC without a manifest?) that must also be resolved before any of this.

iherman:
@HadrienGardeur as I said, I fine postponing the discussion but not closing it. Maybe I am naïve, but I do not see such major issues, because I am also fine restricting the usable versions of TOC to the least complex ones.

Also: I would also be fine saying that a WP-ness _requires_ the presence of a manifest. If its only role, in a specific case, is to ensure the declaration of being a WP, that is fine with me. What I want to avoid is that, in simple cases, the author would have to repeat the same information several times: I do not think that would catch on.

But again: postponement if fine.

HadrienGardeur:
> Also: I would also be fine saying that a WP-ness requires the presence of a manifest. If its only role, in a specific case, is to ensure the declaration of being a WP, that is fine with me.

I'm not sure that a link to an empty document is a great idea though. Same thing for an empty script element.

BigBlueHat:
@HadrienGardeur this has been mentioned vaguely a few times: "this is only one example, not all table of contents look the same." from https://github.com/w3c/wpub/pull/46#issuecomment-324623834

Would you be able to find examples (print, digital, or whatever) that would present something unique or (more importantly) *preventative* from deducing a reading order?

The more concrete the examples, the more concrete the spec.

HadrienGardeur:
I've seen such publications before but publishers like @laudrain might have an easier time than me providing such examples quickly.

HadrienGardeur:
I'll also cc @llemeurfr and @JayPanoz in here, they might be able to contact content producers to obtain examples for:

- publications where the table of contents points to non-linear resources in EPUB
- publications where the TOC has a hierarchy that differs from the reading order

It's not hard to find publication that skip primary resources, there are a lot of examples out there in EPUB.

I know that @baldurbjarnason has provided examples for non-linear publications before, these might also be relevant here.

TzviyaSiegman:
@BigBlueHat I agree with @HadrienGardeur that many authors and publishers craft specialized ToCs, such as cookbooks with pointers to important recipes (but not every recipe) or travel guide ToC with links to Points of interest and important maps (and many other examples). But that should not prevent us from writing a fallback, which can be a generic ToC instance that should be invoked when the ToC is not supplied or it fails for some reason.

JayPanoz:
> specialized ToCs, such as cookbooks with pointers to important recipes (but not every recipe) or travel guide ToC with links to Points of interest and important maps (and many other examples).

Yeah, I was about to list those examples. Besides, it can impact other nonfiction publications, I can remember exploring the idea of a user-centric TOC for a publisher, so that “readers could prioritize the parts they want to read first.”

If “how people actually read” is also something of a concern, @arhomberg can probably offer interesting insights, especially for nonfiction.

JayPanoz:
Also had a book for which that was the real table of contents, and the EPUB one a simplified version. I’ll explain. 

destination

For non french speakers, “Choisissez votre destination” means “Choose your destination.” It was written as a digital-first publication and can’t work at all in the linear model, there’s simply no start nor end; there are two parallel paths, which are unordered lists.

I think that’s worth mentioning, since the concepts of linearity and a reading order (even a default one) has been one of the major complaints publishers expressed when I worked on digital-only projects in EPUB. I know we’re talking about a fallback there, so it’s a FWIW.

arhomberg:
Hi @HadrienGardeur, 

Could you describe one or more examples where the primary reading order/sequence (as _intended_ by author/publisher/creator) would **not** be the same as the TOC?

Obviously, users do NOT always read in the intended sequence (skipping, jumping forward/backwards are all reasonably common behaviours) but in what circumstances would the TOC not reflect the _intended_ sequence of chapters/reading sequence? Even for nonfiction the most common behaviour for readers is to follow the sequence as intended/presneted as opposed to "cherry picking". In fact ebooks reduce the probability of people jumping in and out due to the restrictions the format de factor imposes.

I think your concerns would be easier to understand with some examples/edge cases. :)

arhomberg:
@JayPanoz would it be fair to describe your example as an "enhanced" ebook or at least a "branched" book - a fairly "advanced" use case for 99% of publishers (both tarde and non-trade)

BigBlueHat:
Thanks, @TzviyaSiegman. :smiley:

Thanks everyone for the examples! Sorry I requested them here... Let's do move them (and add more!) to the [ToC-Samples wiki page](https://github.com/w3c/wpub/wiki/ToC-Samples).

...we now return this issue to it's actually intended use. 😁

iherman:
If the TOC is "non-linear", then we _do_ need the reading order explicitly in the manifest. What this tells me, and I believe we can agree on that, that the manifest MUST have a slot for a default reading order, and we cannot rely exclusively on the TOC in one of the resources. This also means that the authors of, say, cookbooks or travelbooks as WP-s should be aware of that, and they MUST provide an explicit reading order in the manifest.

However. The question is what is the percentage of such Web Publications among all Web Publications. Although we do not necessarily have empirical evidence, I believe that most of the WP-s would be much more "regular", whereby I mean that TOC entries follow the regular reading order, although they would refine that greatly, providing, e.g., links to individual sections within some of the resources. I would even go as far as saying that the vast majority of WP-s would be like that.

Taking the vast majority into account, a _fallback_ on those continues to make sense to me, I must say. After all, fallback means "use this if the authoritative mechanism is not provided", where the authoritative mechanism is to have the default order in the manifest.

iherman:
> Thanks everyone for the examples! Sorry I requested them here... Let's do move them (and add more!) to the ToC-Samples wiki page <https://github.com/w3c/wpub/wiki/ToC-Samples>.

I am sure this collection will become extremely useful later, so if we have more, let us have them here indeed!

mattgarrish:
> where the authoritative mechanism is to have the default order in the manifest

To repeat myself, I'm sure, but can't we make a distinction between authoring intent and fallback processing?

For example: The reading order must be included in the manifest, which is either an ordered list of primary resources or a link to an html nav element that contains such a list of links (needing these to be clearly different links, of course).

These are the accepted ways to provide a reading order, but a reading system could piece one together by searching for a toc nav, inspecting the list of resources for certain media types (all html documents), etc.

iherman:
@mattgarrish,

> For example: The reading order must be included in the manifest, which is either an ordered list of primary resources or a link to an html nav element that contains such a list of links (needing these to be clearly different links, of course).
>
> These are the accepted ways to provide a reading order, but a reading system could piece one together by searching for a toc nav, inspecting the list of resources for certain media types (all html documents), etc.

I may have missed this before, but I actually like this approach. In other words, at least for reading order but also for the TOC (and maybe for other things) we do something more explicit like this. It is also more efficient I suppose. Ie,

* The manifest SHOULD include a TOC, which may either be a list of links, or a link to a `nav` element in a primary resource (with the proper `role`)
* The manifest MUST include a default reading order, which again may be either a list of links or a link to a `nav` element in a primary resource (with the proper `role`); in this case the extra rule is the the links extracted from the `nav` element would be pruned to remove fragment IDs and duplicate entries
* The manifest MUST include a title, which is either a string or a link to a `title` element.

And, for each case the fallback is not normatively defined but left it to the UA (using the text you have already here and there).

The extra load on the author may then become minimal. To use the HTML5 Spec example, it would involve a small JSON file with 2-3 links. No big deal (as opposed to repeat the TOC, for example).

My first reaction is: I think I like that.

I am not sure about the language, though...

HadrienGardeur:
@iherman 

> The manifest SHOULD include a TOC, which may either be a list of links, or a link to a nav element in a primary resource (with the proper role)

The TOC could also be a secondary resource IMO, it doesn't have to be in the reading order of the publication.

mattgarrish:
Primary resource is decoupled from reading order (at least for now). Primary is one that is not nested within another (top-level), so it would account for a toc outside the reading order.

HadrienGardeur:
@mattgarrish well I missed the part when primary resources became decoupled from the reading order...

I'm not a fan of talking about nesting, I would much rather have:

- primary resources = reading order
- secondary resources = everything else that's within the boundaries of the publication
- external resources = everything referenced by primary or secondary resources that won't be packaged or cached for offline reading

HadrienGardeur:
@iherman 

> The manifest MUST include a default reading order, which again may be either a list of links or a link to a nav element in a primary resource (with the proper role); in this case the extra rule is the the links extracted from the nav element would be pruned to remove fragment IDs and duplicate entries

I'm still unhappy about that. Aside from TOCs that do not follow the reading order, reference fragments and repeat references to primary resources, there are other issues that are not addressed here:

- a TOC pointing to secondary resources (which is the equivalent of pointing to a non-linear document in EPUB)
- resources in the reading order that are not referenced by the TOC (usually preliminary contents or very large chapters/sections that are divided into multiple resources)

There are too many situations where a TOC does not contain the reading order, I really don't think it can be trusted as a fallback.

mattgarrish:
It was in the last PR. We had two problems with the definitions: one is that primary being reading order and reading order being primary is circular. The other, as we discussed in another thread, is that secondary is tied to being part of the rendering of a primary.

If primary are only resources in the reading order, there's almost no point in having a distinction. There are just resources of which some are in the reading order. The purpose of the rest is indeterminate until encountered. I just don't like this as far as any instruction about what needs to be in the manifest v. what is optional. Eventually it has to come around to some interpretation of standalone resource v. helper resource.

But this PR probably isn't the best place to discuss.

BillKasdorf:
We also need to remember to qualify this as being about WPs that consist of more than one primary resource. As Benjamin reminded us, scholarly journal articles will surely be WPs and most consist of a single primary document, for which reading order and TOC are meaningless. There are millions of those out there. . . .

HadrienGardeur:
@BillKasdorf 

Such a publication would have a single resource in the list of primary resources: problem solved.

BillKasdorf:
Exactly. List of primary resources, always needed. Reading order or TOC, not always.

HadrienGardeur:
@mattgarrish probably not the best place, but this is completely tied to our discussion here...

If you follow my suggestion for primary vs secondary vs external, then secondary resources are useful for:

- packaging
- caching

I really dislike the notion of separating primary resources from the reading order, I don't see the point aside from making things more complex than they should be.

For secondary resources, treating them as "sub" or "nested" resources is fairly useless without knowing which resource reference them. We'd be better off just listing all resources that are not in the reading order.

llemeurfr:
This fork of the initial issue topic should IMO be moved to #16 (non-linear resources - primary, secondary or something else), where is seems it belongs.

iherman:
I must say I am at loss, @HadrienGardeur, because I do not see the problem. If, for whatever reasons, the TOC in a resource is _not_ appropriate for that purpose, then the author can (and should) decouple it from the reading order. In the new scheme (thanks to @mattgarrish) the author would use the manifest in its full beauty to define the reading order. The only extra feature provided is that _if_ the TOC is fine, then a pointer to the TOC suffices. 

In other word, for it to be valid, the manifest MUST include an explicit default reading order. The only think is that the value of that default reading order may be a link to a TOC rather than a list of resources. It is (it must be) a conscious decision of the author (no automatic fallback).

mattgarrish:
@HadrienGardeur Yes, I can still live with where we ended up in #16. Nothing in the world is going to make someone list a major content resource omitted from the reading order, whatever we call it. That was the distinction I was trying to eke out is that some resources are still more important than others, even if not listed.

mattgarrish:
If the author explicitly links to a toc, then they've made the decision it is okay. That's how I can live with that.

As far as fallbacks, I don't think we should ever mandate an algorithm. Finding a toc nav in a primary resource and using it is just one option of what a UA might choose to do. Same with how title/language are worded.

HadrienGardeur:
Fully agree with @mattgarrish.

GarthConboy:
It does sounds like we're now on roughly the same, er, page here -- excellent.

And, just to provide an explicit thumbs up to a comment from @HadrienGardeur above:

> primary resources = reading order
> secondary resources = everything else that's within the boundaries of the publication
> external resources = everything referenced by primary or secondary resources that won't be packaged or cached for offline reading

If an author _explicitly_ references a <nav> element as the reading order (from the manifest), so be it, and those resources (with requisite trimming) are thus the primary ones.

BillKasdorf:
Okay by me as long as we remember to say _default_ reading order.

mattgarrish:

PR #47, opened 2017-08-24
Retrieving a TOC from HTML files

iherman:
This is a spin-off from #46, containing the proposed changes on finding a TOC in HTML files if not explicit in the manifest.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/toc-from-html.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/9718f6a...b30529c.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -427,23 +427,24 @@
 			<section id="table-of-contents">
 				<h3>Table of Contents</h3>
 
-				<p>If the table of contents is not specified in the <a href="#manifest">manifest</a>, the user agent MUST make an attempt to retrive one as follows:</p>
 
-				<ol>
-					<li>
-						Check, among the <a>primary resources</a> of the publication, an HTML resource with basename <code>index.html</code> if it exists, and extract a <a>TOC navigation element</a> from the resource if present;
-					</li>
-					<li>
-						otherwise, locate the first HTML resource among the <a>primary resources</a> in the order provided by the <a>default reading order</a> that also includes a <a>TOC navigation element</a>, and extract it.
-					</li>
-				</ol>
-
-				<p>If such <a>TOC navigation element</a> is found, its content provides the table of contents information for the Web Publication infoset. If several of those are found, only the first one is considered.</p>
-
+				<p>The table of contents is recommended in the infoset.</p>
+				
 				<p>
-					A <dfn>TOC navigation element</dfn> is an HTML <code>nav</code> element with the <code>role</code> attribute set to <code>doc-toc</code>&nbsp;[[!dpub-aria-1.0]].</p>
+					An author MAY omit the table of contents from the <a href="#manifest">manifest</a> if it is included in an HTML <code>nav</code> element identified by the role <code>doc-toc</code>&nbsp;[[!dpub-aria-1.0]]. 
+					This <code>nav</code> element MUST be in a file named <code>index.html</code> or in the first HTML resource in the <a>default reading order</a> (if applicable). In this case, the user agent MUST use this navigation element as the table of contents for the Web Publication.
+				</p>
+				
+				<p>
+					If a user agent requires a table of contents and one is not specified, it MAY provide one of its own making. This specification does not mandate how such a table of contents is created. The user agent might:
 				</p>
 
+				<ol>
+					<li>attempt to locate table of contents nav element in a subsequent primary resource in the default reading order;</li>
+					<li>use the titles of the primary resources in the default reading order;</li>
+					<li>calculate a table of contents using its own algorithms.</li>
+				</ol>
+				
 				<p class=note>
 					This process may not yield a table of contents, which is in line with the fact that a table of contents is not a required information in the <a href="#infoset-req">WP infoset</a>.
 				</p>
--- a/index.html
+++ b/index.html
@@ -421,12 +421,47 @@
 				<h3>Default Reading Order</h3>
 
 				<p>The default reading order MUST include at least one <a>primary resource</a>.</p>
+				
 			</section>
 
 			<section id="table-of-contents">
 				<h3>Table of Contents</h3>
 
-				<div class="ednote">Placeholder for identifying table of contents - whether embedded or by reference.</div>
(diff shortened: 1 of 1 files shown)


iherman:
The current text refers to a `index.html` file as a "default" place for a TOC. I am a bit averse fixing such a name, although I recognize that using `index.html` is fairly widespread on the Web...

BigBlueHat:
Yeah. No magic strings, please. 😃 

Some of the language in this and #46 both talk about "checking the primary resources" of the publication. It seems entirely likely (especially in scholarly publication) that there may be only one primary resource and that itself may be *the* Web Publication--no ToC (unless its internal references) and no "reading order" (per se..."scroll down" 😁 ).

There's probably some exploration around discovery and when any of this "Web Publication-ness" actually has meaning and value.

If HTML is used as the manifest format (or contains it inline), then the ToC and reading order (however constructed/stated) will be "on hand" or it could be considered unnecessary--implicit, single document Web Publication (ala scholarly article).

If the manifest is external and not the primary document requested, then we'll have a two-trip look-up to determine that the document we have is the only one we want/need.

In the multiple document publications there's likely to be *either* a Table of Contents or a "reading order" list--which if you squint looks a lot like a poorly written Table of Contents. 😉 

Would it make sense to flow chart these fallback options? As well as the internal/external manifest look-up "paths" we're tossing around? Those could be discuss-able/presentable sans "serialization" discussions.

Thanks for writing these up, @iherman!

iherman:
I reformulated based on @mattgarrish's formulation in https://github.com/w3c/wpub/issues/48#issuecomment-324613434

index.html
```diff
@@ -421,12 +421,48 @@
...
 
 			<section id="table-of-contents">
 				<h3>Table of Contents</h3>
 
-				<div class="ednote">Placeholder for identifying table of contents - whether embedded or by reference.</div>
+
+				<p>The table of contents is recommended in the infoset.</p>
+				
+				<p>
+					An author MAY omit the table of contents from the <a href="#manifest">manifest</a> if it is included in an HTML <code>nav</code> element identified by the role <code>doc-toc</code>&nbsp;[[!dpub-aria-1.0]].
```
TzviyaSiegman: I think we are asking for trouble if we tie anything to a specific filename.

HadrienGardeur: In this case, the manifest should identify which file is a TOC at least.

iherman: I agree with @HadrienGardeur. I am not enamored with `index.html` either, but the alternative is to get into some other mechanisms (akin the Apache settings that leads to `index.html`) to identify which one is the TOC. We could call it `nav.html` like in EPUB...:-) but `index.html` is the de-facto standard on the Web...

iherman: See also https://github.com/w3c/wpub/issues/48#issuecomment-324671701

HadrienGardeur: The processing rules should be the following:

1. Look for a TOC in the manifest
2. If we cannot find a TOC in the manifest, attempt to identify a primary or secondary resource (based on a rel, or something equivalent in our syntax)
3. If we cannot identify a primary or secondary resource we could then trigger a failure mode:
    1. scan all resources for a `nav@role="doc-toc"` or a `link@rel="contents"`
    2. or default to a well-known location (I'm not a fan)
    3. or extract an outline using HTML elements

TzviyaSiegman: EPUB didn't require any specialized file naming. The TOC was identified by `<nav epub:type="toc">`. This is the same idea as using  `nav@role="doc-toc"` or a `link@rel="contents"`.  Using the combination of these two (one in the file and one in the manifest might be a solution).

HadrienGardeur: With a Readium-like syntax for the manifest, this is what it could look like:

```json
"primary": [
  {"rel": "contents", "href": "/whatever-we-want", "type": "text/html"}
]
```

GarthConboy:
I'm okay with this change -- pending the "index.html" specifics removal as discussed.

iherman:
The comment of @mattgarrish https://github.com/w3c/wpub/pull/46#issuecomment-324899335 may give a twist here, see also my reaction https://github.com/w3c/wpub/pull/46#issuecomment-324907871. To extract from that comment the relevant part related to TOC:

> * The manifest SHOULD include a TOC, which may either be a list of links, or a link to a nav element in a primary resource (with the proper role).

There may be some additional text describing the current mechanism on how the UA MAY extract a TOC, but that can be much looser.

@TzviyaSiegman @HadrienGardeur @GarthConboy ?

HadrienGardeur:
@iherman 

I agree with @mattgarrish suggestion for the TOC as long as we also allow it to be a secondary resource.

Regarding terminology, things are a little fuzzy right now since we haven't decided anything related to the manifest:

- do we talk about links for the items listed in primary or secondary resources?
- do we talk about a role, a relation or a property on such links?

iherman:
@HadrienGardeur I do not understand what you mean. Secondary resources are files like JS files and CSS files, or images; how would they come into the picture here?

HadrienGardeur:
@iherman I disagree.

In a non-linear publication, we could have:

- a single primary resource
- everything else as secondary (including HTML resources)

... if we follow my suggestion for primary vs secondary vs external.

Knowing that a secondary resource is used by a primary resource is only useful when you know which primary resource exactly (this way you can use rel="preload").

I very strongly feel that the current distinction between primary vs secondary is misguided.

iherman:
Fine, I understand. But I am not sure how it influences the mechanism in this issue, ie, that the manifest may have a TOC entry, which is either a list of resources, or a link to a `nav` element in a WP resource.

HadrienGardeur:
@iherman you've just rephrased your previous proposal by saying "a link to a nav element in a WP resource" instead of "a link to a nav element in a primary resource".

If you don't specify primary, I'm OK with the proposal.

iherman:
> 
> @iherman <https://github.com/iherman> you've just rephrased your previous proposal by saying "a link to a nav element in a WP resource" instead of "a link to a nav element in a primary resource".
> 
> If you don't specify primary, I'm OK with the proposal.
> 
> 

Ah! O.k, fine with me.

mattgarrish:

Issue #48, opened 2017-08-24
separate implicit information from failure handling


mattgarrish:
I find the current approach of making publications valid by having user agents fill in holes worrisome. It will make validation for authors that much harder, because there will be no indication that information is missing, or clarity of what it is.

Can we start being more specific about what information an author can intentionally omit, in other words?

Reliance on implicit computation will never be perfect, but as an example here's how the title section could read:

> The Web Publication's infoset requires a title.
> 
>  An author MAY omit the title from the manifest only when the first resource in the default reading order contains a non-empty <code>title</code> element (e.g., an HTML or SVG document). In this case, the user agent MUST use this title as the title of the Web Publication.
> 
> In the case of an invalid Web Publication that does not contain a non-empty title, the user agent MUST provide one. This specification does not mandate how the title is computed. The user agent might:
> 
> - use the non-empty title of a subsequent primary resource in the default reading order;
> - provide a language-specific placeholder title (e.g., 'Untitled Publication');
> - use the URL of the manifest;
> - calculate a title using its own algorithms.

(Personally, I'd rather a title only be implied if there is exactly one resource in the default reading order, but can live with taking the title from the first resource.)

iherman:
@mattgarrish I am fine with this type of reformulation of, e.g., title.

The TOC approach of using `nav` elements is a bit different, though, because the TOC is _not_ a required infoset entry...

lrosenthol:
On Thu, Aug 24, 2017 at 7:04 AM, Matt Garrish <notifications@github.com>
wrote:

> I find the current approach of making publications valid by having user
> agents fill in holes worrisome.
>
Why?  It's what the rest of the web does today (not to mention many other
technologies).


> It will make validation for authors that much harder, because there will
> be no indication that information is missing, or clarity of what it is.
>
> Authors don't validate.  And even if they do, then an actual validator can
report on such missing items (either as errors or warnings).


> Can we start being more specific about what information an author can
> intentionally omit, in other words?
>
Why are we assuming intentional omission vs. simply not having the data
(those are two different things)?  It's omitted - the reason isn't
important.


> An author MAY omit the title from the manifest only when the first
> resource in the default reading order contains a non-empty title element
> (e.g., an HTML or SVG document). In this case, the user agent MUST use this
> title as the title of the Web Publication.
>
> Is "non-empty" the proper terminology?  Do we mean that there is one but
it's an empty/null string?  Or that it's not provided at all?  Or both?


> In the case of an invalid Web Publication
[...]

PR #49, opened 2017-08-24
rewording of title and language

mattgarrish:
Here's an attempt at formalizing the wording from issue #48.

I'm not completely comfortable with it because we're flirting around with serialization, but hopefully it passes muster as a cleanup attempt.

***
[Preview](https://s3.amazonaws.com/pr-preview/mattgarrish/wpub/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/9718f6a...mattgarrish:744c951.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -418,8 +418,8 @@
 				<p>The availability of this address does not preclude the creation and use of other identifiers and/or
 					addresses to retrieve a representation of a <a>Web Publication</a> in whole or part.</p>
 
-				<div class="note">The Web Publication's <a href="#pub-address">address</a> can also be used as value for an
-					identifier link relation [[link-relation]].</div>
+				<div class="note">The Web Publication's address can also be used as value for an identifier link 
+					relation [[link-relation]].</div>
 			</section>
 
 			<section id="wp-resources">
--- a/index.html
+++ b/index.html
@@ -202,6 +202,13 @@
 							and a <a>default reading order</a>.</p>
 					</dd>
 
+					<dt><dfn>Non-empty</dfn></dt>
+					<dd>
+						<p>For the purposes of this specification, non-empty is used to refer to an element or attribute
+							whose text content or value consists of one or more characters after whitespace normalization,
+							where whitespace normalization rules are defined per the host format.</p>
+					</dd>
+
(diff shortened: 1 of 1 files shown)


mattgarrish:
Well, I don't really like harvesting information from other sources, as I've said before, and you're hitting on why. That's why I prefer there be explicit authoring intent for doing so, and preferably this kind of harvesting should only occur when the manifest is embedded in another document where we know this information applies.

Pulling information from random sub-resources will net "something", but the value of that something is in question in my mind. I'm okay with it for failure handling, but we're still playing loose with when an author has done something intentionally.

I did put a note trying to explain that this language has not effect on processing. Originally I had it as a caution, and will update back to that. I'll also try to beef up the explanation that this is informative metadata as far as processing goes.

HadrienGardeur:
I share your concern @iherman and @mattgarrish about publication vs resource level metadata.

We can rely on resource level metadata for the publication as a fallback in failure mode, but we'll need to make it very clear that this is not how publications should be authored.

mattgarrish:
I think what I'll do is change title the same way as language.

I think we're drifting into projecting serialization, and while that has to happen, I don't think we can know how implicit harvesting will work until we have more details.

We can always return to this problem and adjust later.

index.html
```diff
@@ -395,11 +415,11 @@
 				<p>The availability of this address does not preclude the creation and use of other identifiers and/or
 					addresses to retrieve a representation of a <a>Web Publication</a> in whole or part.</p>
 
-				<div class="note">The Web Publication's <a href="#pub-address">address</a> can also be used as value for an
-					identifier link relation [[link-relation]].</div>
+				<div class="note">The Web Publication's address can also be used as value for an identifier link relation
```
HadrienGardeur: If we have a canonical identifier, it probably makes more sense to use the canonical identifier instead of the address in `link@rel="identifier"`.

The `identifier` value is just a proposal at this point, not sure that we should talk about it at all.

mattgarrish: Could you open as a new issue? I'm working on consolidating the three PRs and reverting the defs so will be closing these off shortly.

mattgarrish:

Issue #29, opened 2017-08-14
For manifest in FPWD: Should Natural Language be Required per WCAG 2


Ryladog:
https://rawgit.com/w3c/wpub/manifest-consensus-proposal/index-manifest-proposal.html#abstract-versus-concrete-manifest

lrosenthol:
I assume that the relevant portion of WCAG on this one is https://www.w3.org/TR/2008/REC-WCAG20-20081211/#meaning-doc-lang-id

If so, this clearly refers to a web *page* - which is a single content document in a publication.  There is *nothing* here in WCAG about the requirement that language be specified for the entire publication.

Also, given that the manifest (where the language will appear) does not have any text that needs to be specified - such a value in the manifest serves no practical need.  However, it might be useful as metadata for something like search or identification - but that's why it shoudl be optional.

avneeshsingh:
In WCAG 2.1, a publication consisting of set of files can also fall in definition of Webpage. The example of publication is added in definition of webpage
http://www.w3.org/TR/WCAG21/

This was to provide update for the work done in WCAG 2.1. 
But coming to practical use, AT is more concerned with language of current HTML page instead of the language code placed in manifest. 
Language in metadata is important for search and discovery for accessibility reasons also. If my screen reader does not support Japanese, I should know from metadata itself (before loading the book) that my AT will not work on the publication.

lrosenthol:
On Wed, Aug 16, 2017 at 7:01 AM, Avneesh Singh <notifications@github.com>
wrote:

> Language in metadata is important for search and discovery for
> accessibility reasons also. If my screen reader does not support Japanese,
> I should know from metadata itself (before loading the book) that my AT
> will not work on the publication.
>
I agree with that, Avneesh.  Which is why we should consider the definition
of Language in the metadata with "search and discovery" in mind and not
with the eye towards reading (as it would be on the page).

As such, I see that having two implications
1 - It's optional, because if the author doesn't put it in there, then the
publication won't be found and it's their fault.
2 - We need a structure that allows for multiple values, to enable proper
specification of a multi-lingual publication.

laudrain:
I do agree with 1 and 2.
1 will also be for AT **reading the metadata** (title, authors) in proper language.

iherman:
Thinking about this a bit further, I just realized that

* We know that each HTML file (looking at those only for the moment) *have* a language tag (albeit it may be the default, ie, English)
[...]

PR #51, opened 2017-08-26
cleanup of infoset requirements and fallbacks for title, language, toc and default reading order

mattgarrish:
Looks like the changes got merged and the old PR re-issues in my last attempt. I've reverted and this one's showing the right changes. Apologies for the extra email.

-----------------------------

This PR is a consolidation of the changes in PRs #46, #47 and #49.

Also includes a few additional reversions/changes to the terminology that arose, stemming from issue #16:

- reverting the definition of primary resource to a resource in the default reading order
- changing secondary resource from required for a primary resource to required for the publication
- moving default reading order definition inline into its section, otherwise it creates duplication/redundancy

Please give this a good look over to make sure I didn't mistranslate the discussions.

***
[Preview](https://s3.amazonaws.com/pr-preview/mattgarrish/wpub/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/9d0ed1c...mattgarrish:5dbb9c9.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -348,25 +348,23 @@
 			<section id="wp-language">
 				<h3>Language</h3>
 
-				<p>The language specified in the Web Publication's infoset identifies the natural language of its
+				<p>The language specified in the Web Publication's infoset identifies the natural language(s) of its
 					content.</p>
 
-				<p>This language is not used in the processing or rendering of the Web Publication, and is
-						<strong>not</strong> a replacement for identifying the language of each resource as defined by its
+				<p>This language is not used in the processing or rendering of the Web Publication (including the manifest),
+					and is <strong>not</strong> a replacement for identifying the language of each resource as defined by its
 					format. It instead allows a user agent to ability to provide supplementary enhancements, such as the
 					ability to download a custom dictionary or the preload a language-specific text-to-speech module.</p>
 
 				<p>When specified, the language MUST be a tag that conforms to [[!BCP47]] or its successors.</p>
-
-				<p>In the case of a multilingual Web Publication, the language declaration can be repeated.</p>
 
 				<p>If a user agent requires the language and one is not available in the infoset, it MAY attempt to determine
 					the language. This specification does not mandate how such a language tag is created. The user agent
 					might:</p>
 
 				<ul>
-					<li>use the <a>non-empty</a> language declaration of the first primary resource in the default reading
-						order that provides one;</li>
+					<li>use the <a>non-empty</a> language declaration of the manifest;</li>
+					<li>use the first non-empty language declaration for a <a>primary resource</a>;</li>
 					<li>calculate the language using its own algorithm.</li>
 				</ul>
 
@@ -386,10 +384,12 @@
 					It SHOULD be an <a href="#pub-address">address</a>, but, if not, it MUST be possible to make a one-to-one
 					mapping to an address.</p>
 
-				<p>If the canonical identifier is a URL, it MAY be used as the <code>href</code> value of a "canonical"
-					link&#160;[[!rfc6596]] for the Web Publication in its <a>manifest</a> or in the HTTP response header.</p>
-
 				<p>If assigned, this canonical identifier MUST be unique to the <a>Web Publication</a>.</p>
+
+				<p class="note">If the canonical identifier is a URL, it can be used as the target of a "canonical"
+					link&#160;[[!rfc6596]] (e.g., a [[html]] <code>link</code> element whose <code>rel</code> attribute has
+					the value <code>canonical</code> or a Link HTTP header field [[rfc5988]] similarly identified).</p>
+
 			</section>
 
 			<section id="wp-address">
--- a/index.html
+++ b/index.html
@@ -171,7 +171,7 @@
 				<h3>Terminology</h3>
 
 				<p>Wherever appropriate, this document relies on terminology defined by the note on "Publishing and Linking
-					on the Web" [[!publishing-linking]]. In particular, for the following terms: <a class="externalDFN"
+					on the Web" [[!publishing-linking]], including, in particular, <a class="externalDFN"
 						href="https://www.w3.org/TR/publishing-linking/#dfn-user">user</a>, <a class="externalDFN"
 						href="https://www.w3.org/TR/publishing-linking/#dfn-user-agent">user agent</a>, <a
 						class="externalDFN" href="https://www.w3.org/TR/publishing-linking/#dfn-browser">browser</a>, and <a
@@ -241,7 +241,7 @@
 
 				<ul>
 					<li>It MUST have a <a>manifest</a> that conforms to all normative requirements defined in <a
-							href="#manifest-conformance">Manifest</a>.</li>
+							href="#manifest-conformance"></a>.</li>
 				</ul>
 
 				<p id="ua-conformance">A user agent is conformant to this specification if it meets the following
@@ -269,9 +269,9 @@
 					concrete when the user agent creates an internal representation of the information.</p>
 
 				<p>The infoset does not require a specific serialization. It is primarily compiled from a Web Publication's
-						<a>manifest</a>, whose serialization requirements are defined in <a href="#manifest">Manifest</a>. It
-					is therefore possible to express the same infoset via different manifests, although a Web Publication
-					will only have one manifest.</p>
+						<a>manifest</a>, whose serialization requirements are defined in <a href="#manifest-serialization"
+					></a>. It is therefore possible to express the same infoset via different manifests, although a Web
+					Publication will only have one manifest.</p>
 
 				<p>Although the manifest is the primary source of the infoset, some information may be obtained independent
 					of it. For example, fallback rules for properties defined in the following subsections allow a user agent
@@ -367,9 +367,10 @@
 				<ul>
 					<li>use the <a>non-empty</a> language declaration of the first primary resource in the default reading
 						order that provides one;</li>
-					<li>use the value "<code>und</code>" (undetermined);</li>
 					<li>calculate the language using its own algorithm.</li>
 				</ul>
+
+				<p>If a language tag cannot be determined, the value "<code>und</code>" (undetermined) MUST be used.</p>
 
 				<div class="issue" data-number="29">
 					<p>The question is whether the manifest MUST include the language(s) of the content or not.</p>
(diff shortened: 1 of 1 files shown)


iherman:
@mattgarrish bravo:-)

I have some purely editorial comments. I list them here, but is independent of my "review" comment...

* [ ] "1.3 Terminology": this section should be explicitly set to normative (the enclosing section is set to be non-normative...)
* [ ] First paragram in 1.3, sentence "In particular, for the following terms: user, user agent, browser, and address.": somehow the English does not sound right. Maybe simple make it part of the previous sentence?
* [ ] "3.1 Overview", second para, second sentence:  "It is primarily compiled from a Web Publication's manifest, whose serialization requirements are defined in Manifest." It reads funny with the word 'manifest' repeated. Maybe the second occurence should be spelled out as "in a separate section", or something similar (or the relevant section heading should be changed to allow for respec to do its work)
* [ ] At some point it is probably better to reorder the subsections in section 3 to follow the list of required items of 3.2
* [ ] "3.4 Language", end of first para: Maybe it is worth emphasizing that the language tag is also used as the default language tag for other information items or metadata where appropriate, like title, DC descriptions, etc. (Unless overwritten like, for example, if the title is extracted from a resource with its own language setting)
* [ ] "3.5 Canonical identifier", second paragraph, the canonical can also be used, as far as I understand, for an HTTP response header; worth mentioning. Also, the `link` element is an HTML element; just to be precise we may want to add that.

mattgarrish:
Thanks, Ivan, I'll see what I can do about your notes.

Note that 1.3 is normative now after the last PR (informative is now applied to subsections instead of the introduction itself). I took a look at HTML and some other specs and that's the approach they've used for these preliminary mixes of informative and normative.

I keep thinking about reshuffling, but we keep tweaking the lists so I've held off. As that's purely editorial, I'll do after we get this PR out of the way.

Also, would it make also sense to clarify under resources that the primary resources are compiled from the default reading order, since that already is the list of primary resource? Or too early?

mattgarrish:
> Maybe it is worth emphasizing that the language tag is also used as the default language tag for other information items or metadata where appropriate

This is what confused me in the original issue, as EPUB differentiates the xml:lang of the package document from the dc:language declaration(s) of the content, for example.

A resource can only have one active language declaration for its text content, so what do you do in the case of a multilingual publication if this declaration is both the content and the infoset info?

mattgarrish:
And one last comment...

Why is the html link element brought into the canonical identifier section at all? Isn't that serialization-specific? I've changed to the following, but as I missed the calls where this was discussed feel free to point out if I'm missing the point:

> If the canonical identifier is a URL, it may be used as the href value of a "canonical" link [rfc6596] for the Web Publication in its manifest or in the HTTP response header.

iherman:
@mattgarrish : on the normative vs. non-normative: if the overall section is normative by default, that is fine. 

I would not touch the primary resource issue in this PR. Let us se if it is accepted and come back to it.

iherman:
@mattgarrish : the link element is an HTML element; in other syntaxes it may not exist or, worse, can have a different meaning. Hence it is better, imho, to make it precise...

iherman:
@mattgarrish : my feeling is that the language tag is to be meant _only_ for the manifest and the metadata. Ie, it does not have any effect on the individual resources. If we do it otherwise then, although in rare cases, the behaviour of the browser with the resource may be different than when the same resource is 'accessed' via the WP/manifest.

The metadata is affected by the language tag only if it is in the manifest or in a file referred from the manifest. DC entries used as metadata in the individual resources should not; they are treated as if they were used by the User Agent, independently of the WP.

At leat this is my current feeling...

mattgarrish:
> the link element is an HTML element

Yes, but wasn't it proposed that the canonical/self link be in a JSON expression, too?

We say that there must be a canonical identifier, and that this must be part of the infoset, but we don't say how it's expressed or retrieved.

If it can be used as an html link, where is it coming from and where does this html link belong?

Is it the case that it must be _expressed_ as a canonical link element/property/header so that it can be determined?

mattgarrish:
> my feeling is that the language tag is to be meant only for the manifest and the metadata. Ie, it does not have any effect on the individual resources.

Yes, I agree to an extent. The extent being that it has no bearing on any resources, including the manifest.

That's the impression I got when I asked here: https://github.com/w3c/wpub/issues/29#issuecomment-322758807

This language is not used for processing/rendering, only to describe the publication. It's like the content-language header, where you're just specifying the intended audience.

The language of the manifest file (or any primary resource) will often be the language of the publication, but falls apart with multilingual publications. I can't say I have a bilingual document by putting <code>lang="en fr"</code> in an HTML document, for example.

iherman:
>>    the link element is an HTML element
>
> Yes, but wasn't it proposed that the canonical/self link be in a JSON expression, too?
>

I do not recall about that.

> We say that there must be a canonical identifier, and that this must be part of the infoset, but we don't say how it's expressed or retrieved.
>
> If it can be used as an html link, where is it coming from and where does this html link belong?
>
> Is it the case that it must be expressed as a canonical link element/property/header so that it can be determined?

I believe the whole remark about being able to use the identifier in such a link is a note, rather than part of the core text, actually. We do not talk about how the identifier is expressed or retrieved, just as we do not say anything about the way the TOC is expressed (in the manifest)..

iherman:
@mattgarrish,

>>    my feeling is that the language tag is to be meant only for the manifest and the metadata. Ie, it does not have any effect on the individual resources.
>
> Yes, I agree to an extent. The extent being that it has no bearing on any resources, including the manifest.
>

Hm. You made me reaalize that we have three different roles that we MUST somehow cover.

1. General information on the language(s) of the publication used, e.g., to install or access dictionaries and 
2. Language of the manifest/metadata, ie, the language of textual information like the title (in the manifest) or Dublin Core or schema.org items in the attached metadata
3. Language(s) of the individual resources

I guess we agreed that the language information item has no bearing on No. 3; handling that falls back on how the resources do that per the HTML/SVG/etc. specifications. Maybe we should have, actually, _two_ different language information items for Nos. 1 and 2, with the following extra rules:

* if no publication language is explicitly stated, it has the single "und" value
* if the manifest/metadata language is not stated separately, and there is a single publication language, that one is used; otherwise it has the "und" value

In the absence of a publication language, User Agenta MAY reuse the language information of the first primary resource.

Administratively, maybe we should move this thread to issue #29, though, and merge/close the PR (unless there are other objections)

mattgarrish:
> I believe the whole remark about being able to use the identifier in such a link is a note

Oh, okay, that makes a little more sense. I was trying to figure out where this normatively comes into play. I'll re-adjust.

mattgarrish:
> Administratively, maybe we should move this thread to issue #29, though, and merge/close the PR

Sounds like a plan.

I'll merge later tonight if nothing else comes up. We're not striving to be complete at this stage so there's still plenty of time for debate on everything we've done in this PR.

llemeurfr:
Thanks for the work, @mattgarrish.

Some remaining typos, to be treated in the next PR. 

On 2.1, sentence about the manifest, the last word is now missing. Was "Manifest" before. Ivan proposed ... in "a separate section". 

3.3 Title, the title is now optional, but the note about issue 20 states it is required, which is a contradiction.

3.4 language: I'm surprised to find mention of "BCP47 or its successors". BCP47 is a version independent identifier, the current RFC being 5646. And we also find the contradiction between the optional language in the infoset and the required language in the note about issue 29.

mattgarrish:
Thank @llemeurfr 

> On 2.1, sentence about the manifest, the last word is now missing. Was "Manifest" before. Ivan proposed ... in "a separate section".

Yes, I discovered that this morning while checking for more bad links. Respec should insert the section number/name.

> 3.3 Title, the title is now optional, but the note about issue 20 states it is required, which is a contradiction.

Yes, good catch.

> 3.4 language: I'm surprised to find mention of "BCP47 or its successors". BCP47 is a version independent identifier, the current RFC being 5646. And we also find the contradiction between the optional language in the infoset and the required language in the note about issue 29.

Yes, "or its successors" is definitely unnecessary.

I'll have these updated shortly.

HadrienGardeur:
> Yes, but wasn't it proposed that the canonical/self link be in a JSON expression, too?

That's what we have in Readium: https://github.com/readium/webpub-manifest

Even if it's just a note, I don't think that we should recommend using "canonical" for what we call the canonical identifier. This is used quite differently on the Web, and the "identifier" proposal would be a better fit.

HadrienGardeur:
Going back to the reading order:

- let's imagine a manifest with only a TOC specified in its list of primary resources
- since it has something in the manifest, it's not entirely clear what the UA should do with the current spec language

mattgarrish:
> Even if it's just a note, I don't think that we should recommend using "canonical" for what we call the canonical identifier.

Yes, I'm still kind of confused by this. It's not completely wrong, so long as the "canonical" identifier is the URL of the manifest or the resource it is included in, but the note isn't saying that clearly.

llemeurfr:
> the "canonical" identifier is the URL of the manifest or the resource it is included in

Therefore what is the difference btw this canonical id and the address of the Web Publication?

iherman:
>> Even if it's just a note, I don't think that we should recommend using "canonical" for what we call the canonical identifier.
>
> Yes, I'm still kind of confused by this. It's not completely wrong, so long as the "canonical" identifier is the URL of the manifest or the resource it is included in, but the note isn't saying that clearly.

I am not sure about this at all; the only problem I see with the Note is that it restricts the usage of the "canonical" link to the case when the identifier is a URL. The definition of an identifier says that if it is not a URL per se, it must be possible to make a one-to-one mapping to an address, and, I would think, that address should also be acceptable to be used in a `link` element (or `LINK` header).

We do not say whether or not the address would map onto the manifest; as far as I can see this is still an open issue, related to #5 (except that the comments in #5 went all over the place).

HadrienGardeur:
I thought that the canonical identifier was meant for other identifiers, such as DOIs or ISBNs for example.

iherman:
@llemeurfr 
>> the "canonical" identifier is the URL of the manifest or the resource it is included in
>
> Therefore what is the difference btw this canonical id and the address of the Web Publication?

@HadrienGardeur:
> I thought that the canonical identifier was meant for other identifiers, such as DOIs or ISBNs for example.

Exactly. The identifier is indeed the DOIs and friends, and the URL representations thereof (when necessary) is some sort of a canonical "address". That is what should go into the `link` element, imho, and that is different than the address which might change.

The definition of the identifier also says that the ID must provide a way to get to the manifest. This is not the same as saying it is the address of the manifest.

HadrienGardeur:
@iherman 

This is true for DOIs, not so much for ISBNs which are expressed as URNs. You definitely don't want to use `link@rel="canonical"` with a URN.

mattgarrish:
> This is true for DOIs, not so much for ISBNs which are expressed as URNs.

This is where I'm lost. A canonical link provides the preferred address for a resource. It can overlap with a canonical identifier, but does it always?

iherman:
> @iherman <https://github.com/iherman>
> This is true for DOIs, not so much for ISBNs which are expressed as URNs. You definitely don't want to use link@rel="canonical" with a URN.
> 
> 

True. In which case the spec is silent about the canonical link (which is only a note anyway, not a normative statement).

BigBlueHat:
Question. What's the time/space/tech continuum for "canonical identifier" (re: this PR)?

It's currently stated as:
> If assigned, this canonical identifier MUST be unique to the Web Publication .

Given the following publication, what would it's "canonical identifier" be?
https://www.w3.org/TR/html/

iherman:
@BigBlueHat 

> Question. What's the time/space/tech continuum for "canonical identifier" (re: this PR)?
>
>It's currently stated as:
>
> >   If assigned, this canonical identifier MUST be unique to the Web Publication .
>
>Given the following publication, what would it's "canonical identifier" be? https://www.w3.org/TR/html/

The W3C considers `https://www.w3.org/TR/html/` as THE identifier for the HTML standard, and this approach seems to be fine with its constituent community. However, we have to recognize that other communities may not agree, because the W3C short name refers to the latest HTML standard; this is currently HTML5.2, but it may refer, one day, to HTML6 (if ever there is such thing). Policies on other identifiers may decide that such a major new version should receive a different identifier instead of sharing the same one.

What this tells me is that, in my view, is that how identifiers are used by various communities are _not_ to be defined by this Working Group. It goes way beyond our scope. The information set should provide the right slots to store and use identifiers based on the specification we give, but that is where we should stop, and let other organizations and/or communities establish their own rules.

mattgarrish:
As a formality, can I ask that we stop using this closed PR to discuss issues. It's confusing to follow at this point.

Please open new issues for any clarifications/changes you think are necessary.

Issue #55, opened 2017-08-28
What makes primary resources "primary"


mattgarrish:
Given the new definitions for primary and secondary resources, what gives primary resources any primacy?

Being in the default reading order doesn't make them materially any different from top-level content that the user can reach in other ways.

Something like EPUB's distinction of "spine resources" might be less confusing for times when we need to speak specifically about resources in the default reading order.

mattgarrish:
A primary resource is one that is required for the processing or rendering of a Web Publication, so the distinction has nothing really to do with primacy.

Again, I would call them all "Web Publication Resources" and then find some new name for what a resource in the default reading order is called, if we actually need a special name for such a thing.

Outside of the default reading order, it's not clear we'll need to distinguish resources. Plus, if the default reading order does not contain all of the primary content, talking only about the resources in it as though that's true isn't helpful.

HadrienGardeur:
> Outside of the default reading order, it's not clear we'll need to distinguish resources. Plus, if the default reading order does not contain all of the primary content, talking only about the resources in it as though that's true isn't helpful.

Agree that outside of the default reading order, we don't need to distinguish resources and can have HTML/CSS/JS/images/videos/audio all in a single flat list.

I'll repeat what I said before in a PR comment:

- primary = default reading order
- secondary = part of the publication, not in the default reading order, may be packaged or cached for offline viewing
- external = anything else referenced in the publication, not in any list in the manifest, won't be packaged or cached for offline viewing

In Readium, we still use `spine` instead of primary and `resources` instead of secondary.

mattgarrish:
> In Readium, we still use spine instead of primary and resources instead of secondary.

I think this is the safer approach, or at least the less connotation filled one.

To swipe the secondary definition from google:

> coming after, less important than, or resulting from someone or something else that is primary.

If we don't want to go near "spine", at least let's kill off "secondary" and leave it to the resources section to define these:

> The infoset MUST include a list of resources required for the processing or rendering of the Web Publication, although the list is not required to be exhaustive.
> This list MUST include all primary resources in the default reading order.

I'm not a big fan of the latter statement, though, as it returns us to EPUB.

If we need it for the case of HTML parsing avoidance, let's make that a qualifier.... " when the default reading order is a linked table of contents."

Or make it a requirement for the user agent to compile these into the list: "The user agent MUST include all primary resources in this list. These MUST NOT be duplicated in the resource list in the manifest."

HadrienGardeur:
[...]

PR #57, opened 2017-09-01
genericize resource naming

mattgarrish:
Here's a proposal to end #55 that removes the obvious contradictions of the current primary/secondary. It replaces them with nothing but "resources" and mention of the default reading order, when pertinent.

There's nothing yet that necessitates more complex naming, so let's leave the naming be until there is.

***
[Preview](https://s3.amazonaws.com/pr-preview/mattgarrish/wpub/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/ac65fdc...mattgarrish:07a349c.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -189,8 +189,8 @@
 					<dt><dfn data-lt="Manifests">Manifest</dfn></dt>
 					<dd>
 						<p>A manifest represents structured information about a <a>Web Publication</a>, such as informative
-							metadata, a list of all <a data-lt="primary resource">primary</a> and <a>secondary resources</a>,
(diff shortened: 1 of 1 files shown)


danielweck:
Reviewed. Looks good to me.

mattgarrish:
Thanks, @llemeurfr. I believe "first" is supposed to be "table of contents", but after discussing with Ivan we came to the same conclusion to just leave it until it closes. I'm going to push an update that cleans up the issues after this PR, as we have a number in the spec that are now closed.

PR #62, opened 2017-09-06
rewrite abstract to be both narrower in scope and more descriptive

dauwhe:
The abstract is the first thing people might see when they read the WPUB spec. The current abstract is very general, and doesn’t quite state what we’re actually doing in the spec. I’ve written some new language which probably goes too far in the other direction. I’ve modeled this a bit on the [web app manifest](https://www.w3.org/TR/appmanifest/) abstract, which makes clear that the goal is to provide the information necessary for developers and/or user agents to enable new functionality. The hope is that this might assuage the fears of browser vendors, etc. who were worried about the scope of our charter. 

Feel free to brutally edit or reject the PR; I just think it’s useful for us to start discussing this language.

***
[Preview](https://s3.amazonaws.com/pr-preview/dauwhe/wpub/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/96e37f9...dauwhe:dc3cbad.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -88,11 +88,11 @@
 	</head>
 	<body>
 		<section id="abstract">
-			<p>This specification defines a collection of information that describes the structure and metadata of web publications, 
-			so that user agents or developers may create user experiences well-suited to reading publications. 
-			This information includes the default reading order of the primary resources, a list of secondary resources, 
-			and metadata. This is necessary to enable such behaviors as 
-			sequential navigation and offline reading.</p>
+			<p>This specification defines a collection of information that describes the structure of web publications, 
+			so that user agents or developers may create user experiences well-suited to reading publications, 
+			such as sequential navigation and offline reading.
+			This information includes the default reading order, a list of resources, 
+			and publication-wide metadata.</p>
 		</section>
 		<section id="sotd">
 			<p>This first public working draft provides a preliminary outline of a Web Publication. Many details are under
--- a/index.html
+++ b/index.html
@@ -88,8 +88,15 @@
 	</head>
 	<body>
 		<section id="abstract">
-			<p>This specification defines digital publishing based on a fully native representation of publications within
-				the Open Web Platform.</p>
+			<p>This specification defines an infoset that describes the structure and metadata of web publications, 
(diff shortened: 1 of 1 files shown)


RachelComerford:
Could we say: 

"This specification defines a collection of information that describes the structure and metadata of web publications..."

I had to look up "infoset" and the googles were not helpful :(

rdeltour:
I like the direction of having making the abstract a bit less… abstract, so 👍 from me on the principle.

I also wholeheartedly agree with Rachel that "infoset" is not a great term, so I created #63 to debate this very specific (non-critical) issue.

iherman:
I am not fully convinced. The abstract is very strongly focusing on, essentially, the manifest, whereas the document as a whole may (and probably will/should) address other issues that are, at this moment, only section headings in the document. The language is much more readable than what was there, but is it somehow incomplete. I would also completely remove the second paragraph; the serialization issue is really a detail compared to the higher level goal of what the document will define.

dauwhe:
> The abstract is very strongly focusing on, essentially, the manifest, whereas the document as a whole may (and probably will/should) address other issues that are, at this moment, only section headings in the document.

Agreed. My thought was that we could add to the abstract as we figure out what is actually happening with the other sections.

iherman:
@dauwhe 

> Agreed. My thought was that we could add to the abstract as we figure out what is actually happening with the other sections.

Then we are in a violent agreement. Maybe it is worth adding a small editors' note to make that clear.

I would still remove the second paragraph on serialization, though...

mattgarrish:
So what's a primary resource? (And don't say one in the default reading order! ;)

I would just pare that down to "a list of all resources" and a default reading order. Metadata is also repeated in the first and second sentences, so you might want to clarify what it is and what importance it has.

I'd also change "This is" as the lead-in to the last sentence, as it's kind of vague what it refers to.

mattgarrish:
Comments have gone quiet, and given that the previous abstract was a placeholder line from the DPIG doc, I'm going to accept this in. Like all things at this stage, we'll modify as needed as we go.

PR #64, opened 2017-09-11
First pass at integrating the metadata proposal

baldurbjarnason:
The changes are in two places. Firstly the infoset section is extended with an additional subsection on additional metadata that is all optional as well as the role of external metadata files.

Secondly, we needed to make sure the manifest serialisation has a generic linking mechanism as that's going to be used by a lot of the web publication's moving parts.

First pass, so the usual caveats of the placement, structure, and wording being preliminary and quite possibly suboptimal apply. I also wasn't quite sure about what writing style to use. I'm also guessing at the appropriate HTML structure from the context.

The contributors ('we') here are:
+@BorisAnthony
+@hughmcguire
+@laudrain

(I hope I got everybody's GitHub user ids right. Let me know if I didn't.)

***
[Preview](https://s3.amazonaws.com/pr-preview/baldurbjarnason/wpub/metadata-integration-proposal.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/fc59a55...baldurbjarnason:48964ba.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -486,6 +486,29 @@
 				<p class="issue" data-number="39">There is a consensus that a Web Publication must have a reading order and
 					must/should have a table of contents (the main navigation entry point).</p>
 			</section>
+			<section id="wp-metadata">
+				<h3>Additional Metadata</h3>
+				<p>In addition the stated requirements the Web Publication Infoset SHOULD also provide the following publication metadata:</p>
+				<ul>
+					<li>A list of creators and, optionally, their specific roles. E.g. 'author', 'publisher', 'editor', or 'translator'.</li>
+					<li>Publication date</li>
+					<li>Modification date</li>
+					<li>Accessibility metadata</li>
+				</ul>
+				<p>Identifiers and modification dates are an important aid in the distribution and storage of portable formats (past, current, and present) so authors are urged to include this information whenever they can.</p>
+				<p>The infoset MAY also provide:</p>
+				<ul>
+					<li>A subtitle</li>
+					<li>Licensing information</li>
+					<li>Intended audience</li>
+					<li>General subject</li>
+				</ul>
+
+				<p>These items apply to the web publication in its entirety. Metadata for individual content documents should be embedded in or linked from the content documents themselves.</p>
+				<p>The Web Publication Infoset SHOULD NOT include complex, specialised, and industry-specific metadata and authors should limit the metadata included in the manifest to the items above.</p>
+				<p>If the information set does not cover the publication's needs, authors should link to external metadata files in whichever formats and schema are most commonly accepted as the authoritative metadata format for their intended audiences. They can include multiple metadata links in multiple formats, one for each intended audience, if needed.</p>
+				<p>User Agents MAY decide to support metadata extensions, other metadata schemes, or additional metadata properties if they so choose. User Agents must ignore metadata properties that they do not recognise.</p>
+			</section>
 		</section>
 		<section id="manifest">
 			<h2>Manifest</h2>
@@ -531,6 +554,23 @@
 				<p class="issue" data-number="13">If we have a collection of information about a web publication as a whole
 					("manifest") that exists separately from most of the publication's resources, we need to find a way to
 					associate the manifest with the other publication resources.</p>
+			</section>
+
+			<section id="manifest-linking-from">
+				<h3>Linking <em>from</em> a Manifest</h3>
+				<p>The manifest serialization MUST provide a general linking mechanism for defining a relationship between the web publication and other resources on the web as well as the type of those relationships.</p>
+				<p>This mechanism is used in to express many parts of the web publication's infoset, including but not limited to:</p>
+				<ul>
+					<li><a href="#wp-canonical-identifier">Canonical identifier</a></li>
+					<li><a href="#wp-table-of-contents">External Table of Contents</a></li>
+					<li><a href="#wp-metadata">External metadata</a></li>
+				</ul>
+				<p>This linking mechanism may also be used to express other common link structures on the open web. For example:</p>
+				<ul>
+					<li>Distribution feeds such as <a href="http://opds-spec.org/specs/opds-catalog-1-1">OPDS</a> or <a href="https://tools.ietf.org/html/rfc4287">Atom</a></li>
+					<li>Alternate versions of the publication</li>
+					<li><a href="https://www.w3.org/TR/webmention/">Web Mentions</a></li>
+				</ul>
 			</section>
 		</section>
 		<section id="lifecycle">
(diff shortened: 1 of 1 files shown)


iherman:
(Admin: do not worry for now on the 'all checks failed' mention, taking care of this...)

iherman:
@baldurbjarnason,

first of all, thanks.

Three, overall minor, remarks.

1. I am not sure I understand the following:

> This mechanism is used in to express many parts of the web publication's infoset, including but not limited to:
> * Canonical identifier
> * External Table of Contents
> * External metadata

The "external metadata" if clear. "External Table of Content" is less clear, maybe it is worth giving an example of what you mean, and how it differs from the ToC that is part of the Infoset. What is not clear is "Canonical identifier". Do you man something like "Additional Identifiers" beyond the one that is part of the infoset? I can imagine, say, a scholarly publication having both a DOI and an ISBN, in which case the author might use the DOI as the "real" identifier in the infoset and add the ISBN in an external file, or simply reuse the same identifier by attached to the `dc` namespace. But I am not sure that is what you meant and, if yes, it would be worth stating it more clearly.

2. An extra remark that was in your original document and would be worth, imho, being added as, say, an editorial Note, is that external metadata may be expressed in well known vocabularies like schema.org, Dublin Core, or ONIX, and authors should be encouraged to re-use existing vocabularies wherever they can. This is an informal note, ie, not a normative statement, but would be good to "place" things into the right context.

3. A final, editorial remark is that it is probably better, in view of the overall style, to put Web Mentions, Atom, or OPDS into the reference list, rather then (or additionally) to the link in the document.

Thanks!

HadrienGardeur:
I really like this pull request but have one question to ask to @baldurbjarnason: have you considered support for URI templates for the general linking mechanism?

This would help if we want to link to services (search, dictionary and others) instead of static resources.

mattgarrish:
A few additional editorial comments to Ivan's:

- I'm a bit confused by having "should" requirements in additional metadata. It puts into question how we tie in a pointer from the requirements. The title and language are also only should requirements, so what separates these? If they get that priority, I'd break them out and list them as subsections and leave what remains perhaps as "optional metadata".

- At least as I understand the infoset, it's a generic set of properties that get exposed. At some point, we have to map what a "title" is from the manifest, for example (exact same name, must be a dc:title, something else). Which I only note because I wonder how we can make classes of properties, like accessibility metadata, members of the infoset without naming them. It also raises questions for me about extensibility. Are we dealing with vocab prefixes in the compiled infoset, for example? As I write I see this becoming a separate issue, but I wonder if the infoset is, in fact, nothing more than the processing instructions for the manifest. If we resolved on JSON, as I recall we did a meeting or two ago, we're no longer dealing with consolidating multiple serializations.

- Similarly, the "linking from" section is probably more abstract than it needs to be in a single-serialization world. I'm expecting details like this will be stated explicitly (i.e., where in the json object they occur, how they're named, etc.)

Those nitpicks aside, though, I'm generally fine with the direction of the instructions in this PR.

baldurbjarnason:
I've made a few edits to try to address @iherman's and @HadrienGardeur's comments.

I'm not quite sure what's the best approach for me to integrate @mattgarrish's feedback, though.

I'm a bit hesitant to jump the gun on proposing concrete JSON structures (and if I had to at this point, I'd just crib off what Readium has been doing).

I bundled the pieces in additional metadata together there because they are much less ambiguous than titles, reading order, or language, or whose ambiguity will be largely resolved by the serialisation format (e.g. creator + role) or other groups (e.g. accessibility metadata).

Now, I don't disagree that they might be better off as individual entries (perhaps merging the date parts into one entry) but I'm also a bit limited in the time I can contribute to W3C things (hence this work happening over the weekend). I might not have time to write up the 3-4 additional subsections required until next weekend at the earliest.

iherman:
@baldurbjarnason,

> I've made a few edits to try to address @iherman's and @HadrienGardeur's comments.

Thanks

> I'm a bit hesitant to jump the gun on proposing concrete JSON structures (and if I had to at this point, I'd just crib off what Readium has been doing).

I think we should hold off with that. The issue of serialization is still ahead of us, we do not know, for example, whether we would use JSON-LD or not, whether we would build on top of WAM or not, and these decision would affect the precise serialization.

@mattgarrish,

> I'm a bit confused by having "should" requirements in additional metadata. It puts into question how we tie in a pointer from the requirements. The title and language are also only should requirements, so what separates these? If they get that priority, I'd break them out and list them as subsections and leave what remains perhaps as "optional metadata".

Which may mean that the new list by @baldurbjarnason are MAY and not SHOULD. More exactly: if some are really to be held as SHOULD-s then they should move together with the older list. Looking at the "SHOULD" list f 3.10, the publication and modification dates have more of MAY flair to me, although I am not 100% sure; the list of creators and a11y look more real SHOULD-s, and maybe they could be moved into 3.2 as SHOUld-s

> As I write I see this becoming a separate issue, but I wonder if the infoset is, in fact, nothing more than the processing instructions for the manifest. If we resolved on JSON, as I recall we did a meeting or two ago, we're no longer dealing with consolidating multiple serializations.

I am not sure I agree with this (yet). We do have a number of fallback procedures described in the various sections, meaning that not all of these information items would necessarily appear in a concrete manifest for a specific publication. It is not a matter of multiple _serialization_ per se, but conceptually things are still different.

But we can come back to these later.

iherman:
@baldurbjarnason, just nitpicking: I wonder whether the last paragraph on using URI patterns should be listed in the document as an ISSUE (either there, or adding a separate GitHub issue and making a reference to it); it is a more controversial question. (We already had discussion on this I believe.)

But that can be done as part of a general cleanup

mattgarrish:
> We do have a number of fallback procedures described in the various sections, meaning that not all of these information items would necessarily appear in a concrete manifest for a specific publication.

Yes, but these are steps in processing the manifest to obtain the information, right? If you don't find certain bits of information in the one serialization, go get it from X.

Given that we have an unfilled section for processing the manifest, and that people find the infoset confusing, I just wonder what value it's really providing. If there is only one path for processing the infoset (irregardless of fallback processing), then there doesn't seem to be anything special about it.

But, if we're planning to expose the infoset, my view maybe changes. (But in that case, it needs better definition.)

> I think we should hold off with that.

That's fine with me. I just think longer term that the section on linking from belongs under serialization, since it eventually has to become a concrete part of how the manifest is constructed. I wasn't expecting a rewrite to be specific at this time, only musing that maybe it should be a subsection of serialization.

iherman:
Per [WG Decision](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2017/2017-09-18-minutes.html#resolution2) this will be merged and issues made explicit in the issues' list.

PR #65, opened 2017-09-17
Layout and pagination

frivoal:
This is a first attempt at writing the section about layout and pagination. This is not meant to be right on the first attempt, and will indeed almost certainly be wrong on most points. However, I am hoping this should prove to be a useful starting point to begin discussions and iterations, and to help us move beyond the blank slate.

CC: @TzviyaSiegman @GarthConboy @iherman

***
[Preview](https://s3.amazonaws.com/pr-preview/frivoal/wpub/layout-and-pagination.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/61a5bc2...frivoal:ab80021.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -601,8 +601,8 @@
 				<p class="ednote">Placeholder for inter-publication search.</p>
 			</section>
 
-			<section id="enh-pagination">
-				<h3>Layout and Pagination</h3>
+			<section id="Layout">
+				<h3>Layout</h3>
 				<p>The layout and rendering of <a>Web Publications</a> is governed by the usual rules that apply to all web content.
 				[[!HTML]] (including XHTML) documents are styled and laid out according to the rules of <a href="https://www.w3.org/TR/CSS/">CSS</a>,
 				[[!SVG]] files are rendered as normally expected of that format, and so on.
@@ -713,6 +713,93 @@
 					or of <a href="https://drafts.csswg.org/css-device-adapt/#viewport-meta">the viewport meta tag</a>(see [[CSS-DEVICE-ADAPT]]),
 					or possibly through an extension of <a href="https://drafts.csswg.org/css-page-3/#at-page-rule"><code>@page</code></a> (See [[CSS-PAGE-3]]).
 					This should be discussed with the relevant working groups (<a href="https://www.w3.org/Style/CSS/">CSSWG</a>, <a href="https://www.w3.org/WebPlatform/WG/">WebPlatformWG</a>, <a href="https://whatwg.org/">WHATWG</a>).
+					</div>
+				</section>
+				<section id="enh-pagination">
+					<h4>Paginated Layout</h4>
+
+					<p>When a User Agent renders a <a>Web Publication</a> in a paginated layout,
+					it must lay out each document in the <a>default reading order</a> sequentially,
+					with the last page of a resource being followed by the first page of the subsequent one.</p>
+
+					<div class=issue>
+						<p>To avoid having unstyleable blank pages in the middle of the document,
+						if the preceding resource ended on a left page (resp. right page),
+						the subsequent one should start on a right page (resp. left page)
+						even if the <a href="https://drafts.csswg.org/css-page-3/#progression">page progression</a> (See [[!CSS-PAGE-3]])
+						would otherwise lead to starting on the opposite page;
+						It should also be possible to use the <a href="">break-before</a> property (See [[!CSS-BREAK-3]])
+						to force the content to resume on the opposite side if that was desired by the author.</p>
+
+						<p>[[CSS-PAGE-3] needs to be amended to describe this exception to the general behavior
+						when dealing with collections of documents instead of individual documents.</p>
+					</div>
+
+					<div class=issue>
+						<p>How is that supposed to work when subsequent resources have opposite
+						<a href="https://drafts.csswg.org/css-page-3/#progression">page progression</a> (See [[!CSS-PAGE-3]]) direction,
+						for instance due to different a different writing mode?
+						This is not necessarily a problem from a layout point of view,
+						as each page is independent,
+						but from an UI point of view, if swiping left means next page until the end of
+						one chapter, and starts meaning previous page in the next chapter because
+						we've switched from English to Hebrew, this is going to be horribly confusing.
+					</div>
+
+					<div class=issue>
+						<p>Should any of the above be normative,
+						or should we leave this up to User Agents to explore,
+						and only standardize if and when a clear winning approach emerges?
+					</div>
+
+					<div class=issue>
+						<p>[[CSS-PAGE-3]] needs to be amended so that <a href="https://drafts.csswg.org/css-page-3/#page-based-counters">page counters</a>
+						are not automatically reset to at the beginning of each new resource belonging to the same
+						<a>Web Publication</a>.</p>
+					</div>
+
+					<div class="note">
+						<p>Pagination, like any other aspect of layout, is governed by the various CSS specifications,
+						and is an integrated part of supporting CSS, rather than an isolated standalone feature.
+						Nonetheless, a few specification are especially relevant for User Agents
+						that wish to provide high quality support for paginated media:
+						<dl>
+							<dt><a href="https://www.w3.org/TR/CSS2/page.html">Section in 13</a> in [[CSS2]]
+							<dd>This defines the fundamentals of pagination in CSS.
+							<dt>[[CSS-PAGE-3]]
+							<dd>This CSS module specifies how pages are generated and laid out
+							to hold fragmented content in a paged presentation.
+							It adds functionality for controlling page margins,
+							page size and orientation,
+							and headers and footers,
+							bleed and printer marks,
+							and extends generated content to enable page numbering and running headers / footers.
+							<dt>[[CSS-BREAK-3]]
+							<dd>This module describes the fragmentation model that partitions a flow into pages, columns, or regions.
+							It builds on the Page model module and introduces and defines the fragmentation model.
+							It adds functionality for pagination, breaking variable fragment size and orientation,
+							and controling (forcing or forbidding) where breaks happen, including widows and orphans.
+							<dt>[[MEDIAQ]] and [[MEDIAQUERIES-4]]
+							<dd>These modules provides authors with features to detect various aspect of the media
+							the document is being rendered in, including whether (and how) it is paginated,
+							or more nuanced questions to tell apart environments like a ebook reader from a print rendition.
+							<dt>[[CSS-GCPM-3]]
+							<dd>This module defines a few features that are useful in paged media,
+							such as <a href="https://drafts.csswg.org/css-gcpm/#footnotes">footnotes</a>,
+							<a href="https://drafts.csswg.org/css-gcpm/#running-headers-and-footers">running headers and footers</a>,
+							and <a href="https://drafts.csswg.org/css-gcpm/#the-first-page-pseudo-element">page selectors</a>.
+							<dt>[[CSS-CONTENT-3]]
+							<dd>This module, while overall unstable at the time of writing,
+							contains a few features that while not exclusive to paged media,
+							are particularly useful in paged media,
+							such as <a href="https://drafts.csswg.org/css-content/#leaders">leaders</a> or
+							<a href="https://drafts.csswg.org/css-content/#cross-references">cross references</a>.
+							<dt>[[CSS-PAGE-FLOATS-3]]
+							<dd>This module extends the ability of floats to not only go to the beginning or end of the line,
+							but also to float to the top or bottom of pages, columns…
+							<dt>[[CSS-GRID-1]]
+							<dd>This module is not specific to paginated media, but is nonetheless noteworthy
+							as it enables advanced layouts often found in print publications.
 					</div>
 				</section>
 			</section>
--- a/index.html
+++ b/index.html
@@ -610,7 +610,7 @@
 				other than the expectations set for these technologies by their respective specifications.</p>
 
 				<div class=note>
-					<p>This specification intentially refrains from introducing any new layout feature.
+					<p>This specification intentionally refrains from introducing any new layout feature.
 					Any shortcoming of the web platform in terms of layout needs to be addressed for the whole web platform,
 					which means via CSS.</p>
 
@@ -622,7 +622,7 @@
 					instead of the convergence this specification aims to achieve.</p>
 
 					<p>This working group will work with other relevant groups of the W3C to address
-					plaform wide limitations that negatively impacts <a>Web Publications</a>,
+					platform wide limitations that negatively impacts <a>Web Publications</a>,
 					and invites readers to do the same.</p>
 				</div>
 
@@ -652,13 +652,13 @@
 					<p>An alternative approach would be to leave CSS work on a single document,
 					but have that document be the <a>Web Publication</a> as a whole rather than its individual HTML constituents.
 					This is however merely moving the problem around rather than solving it,
-					as this specification would then need to define semantics for this unified document
+					as this specification would then need to define semantics for this unified document format separately from [[HTML]]
 					including a DOM and assorted APIs, what to do with redundant or conflicting meta data
 					from each each of the constituent part,
 					not to mention a sensible way to merge resources of unrelated types,
 					such as an HTML resource followed in the default reading order by an SVG resource, itself followed by a PNG image…</p>
 
-					<p>However, nothing forces authors to use separate resources for separate chapters of a book,
+					<p>However, nothing forces authors to use separate resources for separate chapters of a book
 					or similar semantic subdivisions of their publications,
 					so this constraint on styling is easily worked around
 					when subdivisions of a publication need to be blended together.</p>
--- a/index.html
+++ b/index.html
@@ -602,9 +602,56 @@
 			</section>
 
 			<section id="enh-pagination">
-				<h3>Pagination</h3>
+				<h3>Layout and Pagination</h3>
+				<section id="scrolling-or-paginating" class=informative>
+					<h4>Scrolling or Paginating</h4>
+					<p>Web content, and a fortiori <a>Web Publications</a>, may be rendered on various media,
+					each with differing characteristics,
+					either due to the physical media itself or to the design of the User Agent.
+					Web Publications are not constrained to any specific medium;
+					User Agents are expected to render them as they would render other web content on the same medium.
+					Authors can adjust the way they style their documents using the features specified in [[MEDIAQ]] or its successor [[MEDIAQUERIES-4]],
+					as supported by the User Agent.</p>
+					<p>It is however worth calling out a major subdivision within media types: scrolling and paginating.</p>
+					<p>In scrolling media, when content overflows the <a href="https://www.w3.org/TR/css-display-3/#initial-containing-block">initial containing block</a> (See [[CSS-DISPLAY-3]])
+					it is exposed by allowing users to scroll to it.
+					This behavior is typical of most desktop or mobile web browsers.</p>
+					<p>In paginating media, content is broken up into discrete pages,
+					and content that overflows one page is displayed on subsequent pages.
+					This behavior is typical of web content printed on paper,
+					of the output of web-to-PDF renderers,
+					and of many e-book reading systems.</p>
+					<p>Some User Agents may support both scrolling and pagination.
+					As the preferences of individual readers vary,
+					and as different types of publications may be better suited for one or the other,
+					this specification encourages User Agents to support both,
+					and to offer a choice to their users.</p>
 
-				<p class="ednote">Placeholder for paginated reading experience.</p>
+					<div class="note">
+					<p>Authors or Publishers occasionally have strong preferences between scrolling and paginating for particular types of content,
+					and may wish to require a publication to always be rendered in one or the other mode.
+					This wish is not, however, something that <a>Web Publication</a> User Agents can effectively enforce.
+					Since <a>Web Publications</a> are composed of backwards-compatible ordinary web content,
+					legacy User Agents unaware of this specification will be able to open them,
+					and will render them as they render any other content.
+					Even <a>Web Publication</a> aware User Agents may only support for one of the two layout modes,
+					and cannot be expected to honor requests to use the other one.
+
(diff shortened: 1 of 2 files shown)


murata0204:
I would like to introduce some issues about the interaction between web publications and HTML or SVG.  In the past, EPUB has arguably encroached the territory of CSS.  All properties mentioned belong are described in the subclauses in [Package Rendering Metadata](http://www.idpf.org/epub/31/spec/epub-packages.html#sec-package-metadata-rendering) of EPUB Packages 3.1. 

Issue: Should web publications indicate whether scroll or "go to the
next/previous page" should be used for HTML primary resources?  (See
the rendition:flow Property)

Issue: Should web publications indicate horizontal centering of HTML and SVG
primary resources?  (see the rendition:align-x-center Property)

Issue: Should web publications indicate whether HTML and SVG primary resources
are pre-paginated or not?  (See the rendition:flow Property)

Issue: Should web publications indicate landscape/portrait/auto?  (See The
rendition:orientation property)

Issue: Should web publications indicate when reading systems should create
the spread view and which HTML and SVG primary resources appear in
left or right pages in the spread?  (See the rendition:spread Property
and the rendition:page-spread-* Properties)

frivoal:
Thanks for these questions @murata0204 , I agree these are the kind of things we need to answer. Here's a quick personal take on these issues:

> Issue: Should web publications indicate whether scroll or "go to the
next/previous page" should be used for HTML primary resources? (See
the rendition:flow Property)

Do you mean choosing between scroll vs go to the next/previous page, or do you mean choosing between scroll/next-page/previous-page vs do nothing? If the former, see next question. If the later, I would say that this should be implyed by the default reading order.

> Issue: Should web publications indicate whether HTML and SVG primary resources
are pre-paginated or not? (See the rendition:flow Property)

This feature should be available, but not specific to Web Publications. My suggestion (included in this PR) is that it should be addressed through an extension of <a href="https://www.w3.org/TR/css-device-adapt-1/#atviewport-rule"><code>@viewport</code></a> or of <a href="https://drafts.csswg.org/css-device-adapt/#viewport-meta">the viewport meta tag</a>, or possibly through an extension of <a href="https://drafts.csswg.org/css-page-3/#at-page-rule"><code>@page</code></a>.

> Should web publications indicate horizontal centering of HTML and SVG
primary resources? (see the rendition:align-x-center Property)

No. There are enough ways in CSS and SVG to center things already. Authors should use just that. And if the problem is poor implementations of these features, that should be solved by fixing the implementations, not by adding more features that will need to be implemented as well anyway.

> Issue: Should web publications indicate landscape/portrait/auto? (See The rendition:orientation property)

This should not be specific to Web Publications. To some degree, this is already available via <a href="https://drafts.csswg.org/css-page-3/#at-page-rule"><code>@page</code></a>. If that is not sufficient, it should be extended, or maybe something should be added to <a href="https://www.w3.org/TR/css-device-adapt-1/#atviewport-rule"><code>@viewport</code></a> or to <a href="https://drafts.csswg.org/css-device-adapt/#viewport-meta">the viewport meta tag</a>. 

> Issue: Should web publications indicate when reading systems should create
the spread view and which HTML and SVG primary resources appear in
left or right pages in the spread? (See the rendition:spread Property
and the rendition:page-spread-* Properties)

Not just for Web Publications. The ability to place things on the right side of the spread should already be available via <a href="https://drafts.csswg.org/css-break/#break-between">the `break-before` and `break-after` properties</a>. The ability to chose between showing pages one by one vs showing both sides of the spread together is currently lacking, and should be added to the web platform as a whole, either via and extension to <a href="https://drafts.csswg.org/css-page-3/#at-page-rule"><code>@page</code></a>, to <a href="https://www.w3.org/TR/css-device-adapt-1/#atviewport-rule"><code>@viewport</code></a> or to <a href="https://drafts.csswg.org/css-device-adapt/#viewport-meta">the viewport meta tag</a>

frivoal:
> I wonder whether the whole of 6.4 shouldn't be marked as non-normative

I think there's one part in there that should be normative: the paragraph that starts with: “For the purposes of layout, each resource of a <a>Web Publication</a> is treated as a separate document.”

and maybe the two paragraph after that, although they could be considered informative explanations of the consequences.

> I  would merge [...], or maybe even [...]

I'd be OK with that change. Should we wait for more opinions, or should I go ahead and do that change?

> ISSUE [...] I would add a different option, namely tht this preference would be expressed as part of the WP's Information set, [...] I would say the same about the portrait/lanscape choice [...]

How about:
1) This is controlled by CSS
2) the default value is auto, which means FOO with ordinary documents and Web Publications that have not expressed any particular preference
3) There is a mechanism in the info set that sets what auto means for all the documents in the Web Publication (but you can still override by setting a different value on each).

I think that'd be OK. At the same time, I wonder if this is really necessary. If that is controllable by CSS, just include a shared stylesheet in all documents, and you'd get the same results. Maybe include a way in the infoset to inject stylesheets into every document while you're at it. Or maybe that's not practical from an authoring standpoint. Either way, I think it is a valid point, but I don't think we should be too hasty in reaching a conclusion.

> If/when we accept this PR, it would probably be a good idea to transfer some of the issues, [...] to the GH issues

Agreed. I didn't want to pollute the issues list with issues about text that hasn't been added yet, but once it is, there should be github issues.

iherman:
> > I wonder whether the whole of 6.4 shouldn't be marked as non-normative
> 
> I think there's one part in there that should be normative: the paragraph that starts with: “For the purposes of layout, each resource of a <a>Web Publication</a> is treated as a separate document.”
> 
> and maybe the two paragraph after that, although they could be considered informative explanations of the consequences.

True enough. Let us then leave it as is for now...

> 
> > I  would merge [...], or maybe even [...]
> 
> I'd be OK with that change. Should we wait for more opinions, or should I go ahead and do that change?

Heh... I am only one commenter:-) But, from my point of view, changing that would only be an editorial change, ie, I think it is fine to go ahead.

> 
> > ISSUE [...] I would add a different option, namely that this preference would be expressed as part of the WP's Information set, [...] I would say the same about the portrait/lanscape choice [...]
> 
> How about:
> 1) This is controlled by CSS
> 2) the default value is auto, which means FOO with ordinary documents and Web Publications that have not expressed any particular preference
> 3) There is a mechanism in the info set that sets what auto means for all the documents in the Web Publication (but you can still override by setting a different value on each).
> 
> I think that'd be OK. At the same time, I wonder if this is really necessary. If that is controllable by CSS, just include a shared stylesheet in all documents, and you'd get the same results. Maybe include a way in the infoset to inject stylesheets into every document while you're at it. Or maybe that's not practical from an authoring standpoint. Either way, I think it is a valid point, but I don't think we should be too hasty in reaching a conclusion.

Actually... the reason we are having, so far, the notion of an infoset is to separate what is the abstract information that should be available from how you would extract it from the sources or put it into a concrete manifest. Ie, it is an infoset item, to be made clear for now, and we may decide later that the only way to have it in a concrete way is to use a CSS included by every document. (I am not sure about any mechanism injecting a style sheet...) 

I believe simply listing that alternative as part of your original text and not going into the details now is good enough. However, maybe we could add what you just proposed as an issue (just that we do not loose it).

> 
> > If/when we accept this PR, it would probably be a good idea to transfer some of the issues, [...] to the GH issues
> 
> Agreed. I didn't want to pollute the issues list with issues about text that hasn't been added yet, but once it is, there should be github issues.

Yep. I can add them later, or you can do it. As you prefer.

Thanks!

murata0204:
@iherman wrote:

> None of these comments are fundamental, they are all minor. They can be handled either pre or post merging the PR.

I think that the grey boundary between publications and CSS is fundamental, and deserves more attention.  To me, this is an area where EPUB has deviated from the OWP.

But I do not think that the FPWD should provide any answers.  I would be happy if it points out this questionable boundary and request feedbacks from Web Browser vendors and other implementors of layout engines, as written in the beginning of this pull request.

TzviyaSiegman:
Thank you @frivoal. I think that we might be able to remove most of the issues and notes because the whole document needs wide review and @dauwhe is working on a section that will address navigation within documents. I think some of the comments belong in best practices (e.g. how to avoid blank pages, which CSS specs are relevant), not in spec. 
So, I think we need to do clean up and shift some items to GitHub issues. Overall, I think this should be fine. I think it would be helpful to hear @dauwhe's comments as well.

frivoal:
@TzviyaSiegman About issues, I'm happy to move them to github / link them from the spec once its merged, as Ivan suggested. Yes, this whole thing will be reviewed anyway, but I would prefer to keep an explicit list of known-to-be-problematic aspects, rather than just hope for people to rediscover them.

As for notes, I generally find specs that explain their consequences, or the motives behind the normative text, to be much easier to read, both for authors and for implementors. While we may disagree about the content or usefulness of the notes, they are not written as "notes to future self" (I use issues for that), but as notes to readers. While deleting them or re-writing them may be appropriate if there's disagreement about their content, shifting them to github wouldn't really work.

Onto specific issues / notes:

* The issue on how to avoid blank pages isn't a problem for authors, it is a problem for spec/implementations: But if we don't say anything at all, what CSS currently specifies is that all documents/resources start on a right page (in the case of horizontal LTR languages like English), so if the last page of a resource was a right one, the first page of the next resource is also a right one, and there's an unavoidable blank in between. I think CSS should be amended to take multi-resource documents into account and to let us have contiguous pages across resources. Maybe I am wrong, but this is something we should consider, and is certainly not just an author best practice.

* Listing CSS specs: This is probably the least necessary note, but I still had a purpose when I wrote it. I want to drive the point that we're not starting from a blank slate, but from the web as currently specified and implemented. I expect that many people in the publishing community are not terribly familiar with the aspects of CSS that are specified but not commonly implemented in mainstream implementations. My assumption this spec will be read by quite a few people who are publishing / layout experts, with a lot of background from the print / paged publication industry, and that they'll want quite a things that aren't commonly available in browsers. My goal here was to help limit the number of feature requests we may get for things that already exist, by pointing out where such things are defined already (for those that are).

Anyway, do you want to merge this, then re-file all the issues as github issues, and possibly open new issues for notes you want to remove, or would you prefer that we discuss each note / issue individually before the merge?

dauwhe:
> For the purposes of layout, each resource of a Web Publication is treated as a separate document.

Our spec says a web publication forms a “single logical work.” And I even have a little toy implementation that does this by creating a single DOM (using HTML imports), where a CSS counter works across resources. 

We already have examples (counters, xrefs) where we don't want each resource treated as a separate document. Do we need to specify this at this early stage?

frivoal:
@dauwhe I think whether to treat documents separately or as a fused document is ultimately the only thing we will need to specify regarding layout, everything else can be handled in CSS. So my point in writing it down is not making sure that everyone agrees with me, but in making sure that we get the debate started on the thing that matters the most.

As for the content of that topic, I think the path you're suggesting is actually much more difficult. In either cases, there will be things to define, but I think that with the approach I am suggesting, new things to be defined will be far fewer and far easier. There may be more than the examples already mentioned, but extending counters and xref to work across an ordered collection of documents instead of on a single document is quite easy, and I think all such things will be limited to small isolated exceptions.

On the other hand, taking the ordered collection of documents and treating it as a single one raises numerous hard questions:
* What do you do with the various things in the `<head>` of the documents? use only the ones from the first document? merge them somehow? For instance, do the stylesheets imported by the `<style>` in the second document apply to the first one as well? Which title does `Document.title` return?
* What does the document's DOM look like. one root with two `<body>`s? or do you fuse the content of every document's `<body>` into a single one? If you fuse the `<body>` elements, what do you do with their attributes? if one body had `lang=fr dir=ltr` and the other one had `lang=ar dir=rtl`, what `lang` and `dir` does the merged body has? If you don't fuse them, what do you do with the various places in HTML and CSS that take attributes/properties of the body (so far assumed to be unique), and use it for the whole document, such as the `dir` attribute, or the background of the body?
* the `tabindex` attribute is set to a document-global number. If you merge two (or more) documents, the scopes collide. How do you sort that out?
* What do you do with `<a href="#foo">...</a>` if several of the merged documents have elements with id "foo"? just pick the first one (changing the target of the link depending on whether the documents are read as a collection or individually), or do you introduce some system that keeps track of which document (before the fusion) the `<a>` element was in? If you do that, how does that tracking survive DOM changes via Javascript?
* same question for `<label for="foo">...</label>`
* Do all sub-documents of the merged document count as visited?
* What if one document was in quirks mode and the other not?
* What do `Document.documentURI ` / `Document.location` / `Document.URL ` return?
* What's the security model when the documents come from different domains? Or is that forbidden?
* Is the `load` event fired once for the whole collection, or multiple times when each sub-document is loaded?
* What happens to namespaces you combine HTML and XHTML documents? Or to selectors, do they match case sensitively or not? Or is this combination forbidden?

I think there's many more like that, and that sorting this out would be a major undertaking. It is not particularly hard to make a toy/naive implementation (vivliostyle has one too), for which the answers to this sort of questions just fall out of the underlying architecture. However, if we want this to be robustly inter-operable, we would need a very long spec, lots of tests, significant efforts on the implementation side, and we would need to keep updating this every time HTML adds a new feature.

iherman:
I think I am with @frivoal on this... (sorry @dauwhe)

Of course, it is possible to have a particular WP that uses the include trick. In which case, as far as I am concerned, it has only one "real" resource. But I do not think that the "include" approach scales. Beyond the issues @frivoal lists, there is simple efficiency issue. Look at the HTML5 specification as an example. By default, it is cut into 15+ resources, but they also provide a single source version. Loading the single source version, even with today's browsers and machines, takes some time; too much for an average reader. Let alone any interaction you have to do with it like scrolling. And this is a pre-generated HTML file, ie, you would not have to have 15+ extra HTTP requests which would be the case if includes were used.

The goal would be to keep the 15+ resources, but provide enough information to let the user agent do, e.g., continuous section numbering, for example.

index.html
```diff
@@ -671,10 +671,241 @@
...
+					and at worst lead to a divergence between “ordinary” web content and <a>Web Publications</a>,
+					instead of the convergence this specification aims to achieve.</p>
+
+					<p>This working group will work with other relevant groups of the W3C to address
+					platform wide limitations that negatively impacts <a>Web Publications</a>,
+					and invites readers to do the same.</p>
+				</div>
+
+				<p>For the purposes of layout, each resource of a <a>Web Publication</a> is treated as a separate document.
+				User Agents must not mix content from multiple resources in the same rendering:
```
lrosenthol: This is not true.  Resources are not the same as documents...

index.html
```diff
@@ -671,10 +671,241 @@
...
+				For instance, <a href="https://www.w3.org/TR/CSS21/visuren.html#floats">CSS floats</a> or <a href="https://www.w3.org/TR/CSS21/visuren.html#absolutely-positioned">Absolutely positioned elements</a>
+				from one resource may not intrude or overlap with content from an other resource.</p>
+
+				<p>Users must therefore rely on the UI features of the User Agent
+				to navigate from one resource to another,
```
lrosenthol: A publication may present that navigation not the UA itself

index.html
```diff
@@ -671,10 +671,241 @@
...
+				rather than expect a collection of resources to be readable as if it were a single one.</p>
+
+				<p>This does not imply that User Agents are forbidden from juxtaposing the rendering of the various resources.
+				Scrolling User Agents may for instance place renderings of the resources beneath each other in the <a>default reading order</a>,
+				so that users may seamlessly scroll past the end of one and into the next one,
```
lrosenthol: Why beneath?  That seems very stringent on implementation

index.html
```diff
@@ -671,10 +671,241 @@
...
+					<p>In paginating media, content is broken up into discrete pages,
+					and content that overflows one page is displayed on subsequent pages.
+					This behavior is typical of web content printed on paper,
+					of the output of web-to-PDF renderers,
+					and of many e-book reading systems.</p>
+					<p>Some User Agents may support both scrolling and pagination.
+					As the preferences of individual readers vary,
+					and as different types of publications may be better suited for one or the other,
+					this specification encourages User Agents to support both,
+					and to offer a choice to their users.</p>
```
lrosenthol: it also needs to be an attribute of a publication as a given author may have a clear position on how they wish their content to be presented

index.html
```diff
@@ -671,10 +671,241 @@
...
+
+					<div class="note">
+					<p>Authors or Publishers occasionally have strong preferences between scrolling and paginating for particular types of content,
+					and may wish to require a publication to always be rendered in one or the other mode.
+					This wish is not, however, something that <a>Web Publication</a> User Agents can effectively enforce.
+					Since <a>Web Publications</a> are composed of backwards-compatible ordinary web content,
```
lrosenthol: I disagree and this is *very* important aspect.  it also needs to be an attribute of a publication as a given author may have a clear position on how they wish their content to be presented

index.html
```diff
@@ -671,10 +671,241 @@
...
+					even if a strict requirement is not possible. This should most likely be addressed through an extension of
+					<a href="https://www.w3.org/TR/css-device-adapt-1/#atviewport-rule"><code>@viewport</code></a>
+					or of <a href="https://drafts.csswg.org/css-device-adapt/#viewport-meta">the viewport meta tag</a>(see [[CSS-DEVICE-ADAPT]]),
+					or possibly through an extension of <a href="https://drafts.csswg.org/css-page-3/#at-page-rule"><code>@page</code></a> (See [[CSS-PAGE-3]]).
+					This should be discussed with the relevant working groups (<a href="https://www.w3.org/Style/CSS/">CSSWG</a>, <a href="https://www.w3.org/WebPlatform/WG/">WebPlatformWG</a>, <a href="https://whatwg.org/">WHATWG</a>).
+					</div>
+				</section>
+				<section id="enh-pagination">
+					<h4>Paginated Layout</h4>
+
```
lrosenthol: I think we should keep this out for now...it's just too controversial

index.html
frivoal: I am using the word document in the CSS / HTML sense. The thing that has a DOM, a document object, and that has a defined model for being styled. Not an abstract concept of document, which could/should be used to refer to the entire publication.

Any suggestion as to how to make that unambiguous?

Now, we may still disagree as to whether that's a good idea (as has been discussed in https://github.com/w3c/wpub/pull/65#issuecomment-330925058 and https://github.com/w3c/wpub/pull/65#issuecomment-331059874 and https://github.com/w3c/wpub/pull/65#issuecomment-331098264), that's a separate question.

index.html
frivoal: Are you saying that UAs should not provide a UI for navigating between the resources?

We should have an API to let publication provide additional or alternative means of dealing with this, but by default I'd expect the UA to provide some means of navigating the table of contents, switching chapters, and so on, otherwise I am not sure in what sense it would be web publication aware.

index.html
frivoal: "may for instance" this is an example, not a constraint. UA innovations is more than welcome here.

index.html
frivoal: Yes, this flag / attribute / css directive should exist. Authors should express their preference and intent.

However, it can only ever be a preference. I do not see how we can possibly come up with a solution that is compatible with legacy HTML UAs (also known as browsers) and at the same time have a mandatory flags that demands that the document is presented in a paginated way. Browsers do not support it today, so effectively, it is already optional. Just the same, even if an author says they want their document to be scrolled, not paged, if you print it, it won't scroll.

In addition to the matter of capabilities, it's a User Agent, not an Author Agent, so if the User and the Author disagree, the User ultimately wins. There are things like user stylesheets, and there will be extensions, and such things can let the user override the author's preference. 

Unless we go about making documents that do not open in browsers, and documents that can be forbidden from printing, making a paged vs scroll flag mandatory to respect by UAs is just impossible.

index.html
dauwhe: I think it's useful to track these issues here at this stage. The web will need to come to terms with pagination, and with "collections" of documents. When our spec is more mature it may make sense to move some of these to CSS, but I think it's too soon for that.

index.html
dauwhe: I'm wary of a flag or other information in a "manifest" that expresses rendering preferences. The publication author gets to write the CSS for the publication, and can express their preferences there.

index.html
frivoal: Would you mind elaborating?

I am not claiming to have the perfect answer, but I think this is a topic we cannot avoid. My goal in writing this section is to say something concrete, even if minimal, so that we can begin working out what it is we agree or disagree about.

As far as I can tell, there are three things we have to have:
* How do you describe a collection of documents/resources
* How do you render the thing
* Where/how do you store the metadata about the collection

If we don't have that, we do not have a specification (although I suppose you could skip the metadata part and still have a specification, just not one we'd be very happy with).

Without a collection and working with a single document/resource, we already know how that works, it's just HTML and CSS. Once you have a collection, we no longer know how the HTML and CSS models apply so we need to state something, and as HTML approaches to metadata do not operate on collections, we need to propose something there as well, lest we have no metadata on the collection.

We cannot possibly get everything right on the first attempt, but I don't think we can in good faith go to FPWD without at least an attempt at answering the question of how a Web Publication renders, including in a paged environment.

frivoal: (to be clear, my comment above was in reply to @lrosenthol)

index.html
frivoal: Right. I think this should be expressed in CSS (through `@page` or `@viewport` or some such, as discussed in the issue below this note). We may also want to have a way to express this for the entire collection, but that should be something that works with CSS. Maybe a way to link to a stylesheet to be shared between all resources in the collection.

dauwhe:
> Scrolling User Agents may, for example, place renderings of the resources beneath each other in the default reading order, so that users may seamlessly scroll past the end of one and into the next one, as long as the layout of resources is processed independently

Does this open up all the complications of having multiple Documents (in the DOM sense) in the same browsing context? Would pieces of two different documents be in the same window as you scroll past the boundary? The paginated case seems simpler, because the browsing context could just navigate to the next document, and page breaks are by definition not seamless :)

I wonder if we should omit this. We're already clear on what happens:

> For the purposes of layout, each resource of a <a>Web Publication</a> is treated as a separate document.

dauwhe:
See #85 for potential edits to this PR.

frivoal:
@dauwhe I don't think this juxtaposition of scrolling content would be particularly problematic. As long as this is a purely graphical operation, with no mixing of document content, it should be alright. That said, this is also a purely speculative UX suggestion, so I have no problem about dropping it.

PS: I'll go read #85

frivoal:
Closed in favor of #85

PR #71, opened 2017-09-21
Add sections on navigation and manifest processing

dauwhe:
I've added small bits about navigation, and more extensive text (mostly copied from web app manifest) on linking to and obtaining a manifest. This is assuming we have an external manifest, which seems likely :)

In some sense, this is an argument for piggybacking on web app manifest, so they get to do all the hard stuff!

This is all very rough, but we should start somewhere.

### DIFF
--- a/index.html
+++ b/index.html
@@ -56,6 +56,7 @@
                       }]
                   },
 			  ],
+			  
               processVersion: 2017,
               includePermalinks: true,
               permalinkEdge:     true,
@@ -571,6 +572,29 @@
 				<h2>Linking to a Manifest</h2>
 
 				<div class="ednote">Placeholder for how resources identify they belong to a publication.</div>
+				
+				<p class="note">This is adapted from the [[Web Application Manifest]] spec.</p>
+
+				
+				<p>A resource is said to be associated with a manifest if the resource 
+				representation, an HTML document, has a manifest link relationship.</p>
+				
+				<p>To link to a manifest, use the [[HTML]] <code>link</code> element 
+				using the keyword <code>manifest</code> as the value of the 
+				<code>rel</code> attribute. This creates an external resource link:</p>
+				
+<div class="example">
+<pre>
+&lt;link rel="manifest" href="manifest.json"&gt;
+</pre>
+</div>
+				
+				<p class="note">In cases where more than one <code>link</code> element 
+				with a manifest link type appears in a Document, the user agent 
+				uses the first <code>link</code> element in tree order 
+				and ignores all subsequent <code>link</code> element with a manifest 
+				link type (even if the first element was erroneous). 
+				See the steps for obtaining a manifest.</p>
 
 				<p class="issue" data-number="13">If we have a collection of information about a web publication as a whole
 					("manifest") that exists separately from most of the publication's resources, we need to find a way to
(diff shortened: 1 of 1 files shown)


mattgarrish:
Well, I have been working on linking to a manifest and the initiation problem, having won that lottery... ;)

There are problems I think we should broach beyond what web app manifest defines, if not in full. These include introducing HTTP linking for non-HTML resources, handling top-level non-HTML resources in general, having to deal with multiple manifests, and that initiation and installation are likely separate activities.

There's also a question of whether the publication is an app (embeds the reading system) or is an external application, and whether both can co-exist.

I'm trying my best not to get too bogged down in details, but it's still hurting my brain immensely. I'm hoping to have prose soon-ish. Hopefully, by Monday, but, if not, shortly after.

dauwhe:
@mattgarrish Ah, I was just trying to get the ball rolling, not realizing you were already rolling it! Feel free to delete the PR, or borrow a sentence or two :)

dauwhe:
> handling top-level non-HTML resources in general,

I only just discovered that the HTML spec does talk about this quite a bit, essentially giving the user agent [responsibility](https://html.spec.whatwg.org/multipage/browsing-the-web.html#read-media) to generate the “missing” elements. 

Lots of interesting questions here!

mattgarrish:
Thanks, I was wondering where that behaviour was defined. I'd seen the HTML5 controls on bare video, but hadn't gone rooting for the source.

Still have a problem of the HTML wrapper only being created during rendering, though. That's where we diverge and need http links, can't have site-provided install/initiate prompts, etc.

baldurbjarnason:
One concern I have, as it's a recurring issue, is making sure we don't needlessly destroy the usefulness of the web publications format to whole swathes of publication genres.

A regular question we get at Rebus from academics is: 'could this be a web publication?' while pointing at a wholly non-linear standalone website (e.g. references, documentation, texts with map-based navigation, proper hypertexts, etc.).

They have to ask because ePub explicitly does not allow such texts without destroying their form, structure, and navigation.

So, to make sure we can stop answering with a vague _'we hope so'_ and instead say 'yes, definitely!' I'd like to propose that we explicitly require that we explicitly require that User Agents do not prevent the web publication from providing own navigation elements.

Something along the lines of this to the Navigation section:

> User Agents MUST NOT prevent regular navigation inside the current reading context from resource to resource within the publication, even when the target resource is not in the specified reading order. Web Publication HTML resources MUST be able to provide their own navigation if that is what best serves their form and readers.

We don't want to be put in the ePub situation where whole genres of publications simply can't be accommodated or where the publication proper can only be experienced within a pop-up window over the main text.

BillKasdorf:
+1 except to suggest, instead of “specified” reading order, saying “default” reading order, which is what I think we’re otherwise saying. I’ve been persuaded that it’s necessary to have a default reading order, but imo there are actually more publications that are consumed _not_ in that default order than those that are, even if they are “published” in an arbitrary order.

Bill Kasdorf

VP and Principal Consultant | Apex CoVantage

p:

734-904-6252  m:   734-904-6252

ISNI: http://isni.org/isni/0000000116490786
ORCiD: https://orcid.org/0000-0001-7002-4786<https://orcid.org/0000-0001-7002-4786?lang=en>

RachelComerford:
+1 to default reading order - it seems to be the common vernacular anyway
which will make it easier for new users to parse.

Rachel Comerford | Director of Content Standards | T 212.576.9433

*Macmillan Learning*

PR #79, opened 2017-10-03
Manifest linking

mattgarrish:
This PR removes section 5 on initiating, based on the discussions on the last call.

It retains a section on linking to a manifest that I wrote while working on 5.  There are issues that need further discussion, of course. It does not disallow linking to multiple manifests, for example, but does not require processing of more than the first link found. The relationship "publication" may also change depending on discussions. It is used to avoid comparisons with Web App Manifest until that issue is fully resolved.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/manifest-linking.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/159950b...f1450f0.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -578,15 +578,47 @@
 			<section id="manifest-linking">
 				<h2>Linking to a Manifest</h2>
 
-				<div class="ednote">Placeholder for how resources identify they belong to a publication.</div>
+				<p>Providing a link from a resource to its manifest allows a user agent to discover that a Web Publication is
+					available. The inclusion of links is not always possible, however. For example, when the resource is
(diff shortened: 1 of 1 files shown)


iherman:
I do not have a problem with the PR per se, but I believe the Wiki page you made on the pros and cons of WAM alignment should be linked from the draft (from a note, or, probably better, from the issue dealing with WAM). As long as that is still an open issue we should not ignore that.

mattgarrish:
Yes, agree, thanks for the reminder. I've added an issue with links to those two.

mattgarrish:
Merging as there have been no additional comments. Feel free to open new issues if you spot anything.

PR #80, opened 2017-10-03
expanded intro

mattgarrish:
This PR adds more detail to the introduction.

It consumes the background section and replaces it with two new sections: one that details why we need web publications and another that details what they are.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/revised-intro.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/159950b...b6fd561.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -114,29 +114,69 @@
 		<section id="intro">
 			<h2>Introduction</h2>
 
-			<section id="background" class="informative">
-				<h3>Background</h3>
-
-				<p>For millenia now, the written word has been the primary means of encoding and sharing ideas and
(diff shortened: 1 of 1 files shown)


dauwhe:
@mattgarrish This is wonderful. Thanks so much for writing so beautifully!

mattgarrish:
Thanks @dauwhe ! Always appreciate a compliment...

iherman:
@mattgarrish I would think: go ahead and merge:-)

mattgarrish:
Sounds like a plan. It's not getting any negative feedback, and it's only informative.

PR #81, opened 2017-10-03
media type registration for manifest file

mattgarrish:
I'll say up front, feel free to nix this. It may be a bit premature.

In any case, in looking at section 5 this was another of the ancillary pieces that looks like it will be needed. It still needs work (agreement on media type and extension, plus security considerations), but didn't want to toss it without offering it first.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/media-reg.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/159950b...4e1f985.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -566,6 +566,14 @@
 				<div class="ednote">Placeholder for how a manifest declares it describes a web publication.</div>
 			</section>
 
+			<section id="manifest-media-type">
+				<h3>Media Type</h3>
+				
+				<p>The media type for a Web Publication manifest is:</p>
+				
+				<pre><code>application/publication+json</code></pre>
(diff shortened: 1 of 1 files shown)


iherman:
I would say: keep it as a PR but I would not merge it yet. Wait until (and if) we have a WG decision _not_ to align with the WAM.

lrosenthol:
Agreed with @iherman - leave this open.  There are ramifications for this
also for PWP as well...

mattgarrish:
Closing this off as it's become stale and we can copy a media type registration into the spec at any time if we decide we need one, which is all this did.

PR #85, opened 2017-10-13
Layout and Pagination (edit to #65)

dauwhe:
Lacking the git-fu to figure out how to edit a PR from a fork, I just made another fork and PR. Sadly, we lose the diff from Florian's version. 

This is very much the content of Florian's PR #65, just with many of the notes removed, along with some other text. The goal was to keep things simple and brief, but to retain the issues. 

I did post a [comment](https://github.com/w3c/wpub/pull/65#issuecomment-336520033) about one of these deletions in #65.

***
[Preview](https://s3.amazonaws.com/pr-preview/dauwhe/wpub/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/b2e6446...dauwhe:da56f8a.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -725,10 +725,113 @@
 				<p class="ednote">Placeholder for inter-publication search.</p>
 			</section>
 
-			<section id="enh-pagination">
-				<h3>Pagination</h3>
+			<section id="Layout">
+				<h3>Layout</h3>
+				
+				<p>Since <a>web publications</a> may consist of many HTML documents considered as a logical whole, this does have implications for layout, as discussed below.</p>
 
-				<p class="ednote">Placeholder for paginated reading experience.</p>
+
+				<p>The layout and rendering of <a>Web Publications</a> is governed by the usual rules that apply to all web content.
+				[[!HTML]] documents are styled and laid out according to the rules of <a href="https://www.w3.org/TR/CSS/">CSS</a>,
+				[[!SVG]] files are rendered as normally expected of that format, and so on.
+				This specification requires no particular profile or subset of CSS, HTML, or SVG to be supported,
+				other than the expectations set for these technologies by their respective specifications.</p>
+				
+
+				<div class=note>
+					<p>This specification intentionally refrains from introducing any new layout feature.
+					Any shortcoming of the web platform in terms of layout needs to be addressed for the whole web platform,
+					which means via CSS.</p>
+
+					<p>This working group will work with other relevant groups of the W3C to address
+					platform-wide limitations that negatively impacts <a>Web Publications</a>,
+					and invites readers to do the same.</p>
+				</div>
+
+				<p>For the purposes of layout, each resource of a <a>Web Publication</a> is treated as a separate document.
+				User Agents must not mix content from multiple resources in the same rendering:
+				For instance, <a href="https://www.w3.org/TR/CSS21/visuren.html#floats">CSS floats</a> or <a href="https://www.w3.org/TR/CSS21/visuren.html#absolutely-positioned">Absolutely positioned elements</a>
+				from one resource may not intrude or overlap with content from an other resource.</p>
+
+					<div class=issue>
+						<p>Despite this general requirement that each resource should be treated as a separate document
+						for the purpose of layout,
+						there are some places where CSS specifications should be amended to
+						be able to deal more intelligently with collections of resources like <a>Web Publications</a>.
+
+						<p>One instance is the definition of <a href="https://drafts.csswg.org/css-content/#cross-references">cross references</a>
+						which are currently restricted to work only within a single document.
+						This restriction should be relaxed to allow for cross references between
+						separate resources of a single <a>Web Publication</a>.
+
+						<p>Another related one would be to allow <a href="https://drafts.csswg.org/css-lists-3/#counter-properties">counters</a>
+						to accumulate across multiple resources of a single <a>Web Publication</a>,
+						for instance so that figures in multiple sections may be numbered in a single sequance.
+
+					</div>
+
+				<section id="scrolling-or-paginating" class=informative>
+					<h4>Scrolling or Paginating</h4>
+					<p>Publications have historically been presented via paged media.
+					Web pages almost always scroll. 
+					As the preferences of individual readers vary,
+					and as different types of publications may be better suited for one or the other,
+					this specification encourages User Agents to support both,
+					and to offer a choice to their users.</p>
+
+					<div class=issue>
+					<p>It may nonetheless be useful for authors to be able to specify a preference between scrolling and pagination,
+					even if a strict requirement is not possible. This should most likely be addressed through an extension of
+					<a href="https://www.w3.org/TR/css-device-adapt-1/#atviewport-rule"><code>@viewport</code></a>
+					or of <a href="https://drafts.csswg.org/css-device-adapt/#viewport-meta">the viewport meta tag</a>(see [[CSS-DEVICE-ADAPT]]),
+					or possibly through an extension of <a href="https://drafts.csswg.org/css-page-3/#at-page-rule"><code>@page</code></a> (See [[CSS-PAGE-3]]).
+					This should be discussed with the relevant working groups (<a href="https://www.w3.org/Style/CSS/">CSSWG</a>, <a href="https://www.w3.org/WebPlatform/WG/">WebPlatformWG</a>, <a href="https://whatwg.org/">WHATWG</a>).
+					</div>
+				</section>
+				<section id="enh-pagination">
+					<h4>Paginated Layout</h4>
+
+					<p>When a User Agent renders a <a>Web Publication</a> in a paginated layout,
+					it must lay out each document in the <a>default reading order</a> sequentially,
+					with the last page of a resource being followed by the first page of the subsequent one.</p>
+
+					<div class=issue>
+						<p>To avoid having unstyleable blank pages in the middle of the document,
+						if the preceding resource ended on a left page (resp. right page),
+						the subsequent one should start on a right page (resp. left page)
+						even if the <a href="https://drafts.csswg.org/css-page-3/#progression">page progression</a> (See [[!CSS-PAGE-3]])
+						would otherwise lead to starting on the opposite page;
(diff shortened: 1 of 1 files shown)


iherman:
I believe this should be merged into the main branch; it contains the essential messages at this point. (Editorially, the issues should be migrated to GH, but that can be done later.)

frivoal:
Here's a diff between your version and mine (differences in other sections to be ignored, as they are simply caused by the base being different): https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fs3.amazonaws.com%2Fpr-preview%2Ffrivoal%2Fwpub%2Flayout-and-pagination.html&doc2=https%3A%2F%2Fs3.amazonaws.com%2Fpr-preview%2Fdauwhe%2Fwpub%2Fmaster.html#Layout

frivoal:
Comments:

* There's a number changes, including a few small ones. For those I do not comment about here, assume that I either agree with the change or if I disagree, that the disagreement is minor enough to be ignored.

* You deleted "Users must therefore rely on the UI features of the User Agent to navigate[...]" and the note that goes along with it. I understand the motivation is to not be overly prescriptive at this point of UA behavior, but I think this goes to far. I was not mandating any particular way the UA needs to make navigation possible, only that it does in some way. If we think that the UA shouldn't be in charge of that, and that this should instead be handled by the document/publication author in javascript, then we need to define some kind of API to make this doable. Otherwise, if there is no UA provided UI nor any API, the whole spec becomes mostly a no-op: while it would offer some meta-data conventions (assuming we do agree on a concrete manifest format), it would offer no no capabilities. I do not believe this was what we set out to do.

* The "Scrolling or Paginating" section has been largely rewritten. While I have no strong attachment to my original prose, which is of course perfectible, I don't particularly agree with the replacement:
  > Publications have historically been presented via paged media.

  Only if we take "publication" in a limited sense. Yes, books and other printed matter, as well as epub content have. If we consider everything on the web to also be a publication (which I think we should), then no. Posters, billboards, flyers, pamphlets, actual scrolls, and various other media types, ancient or modern are not paged either, and qualify just fine for being publications.

  > Web pages almost always scroll

  Web **Browsers** almost always scroll. Web **pages** scroll or page depending on how you look at them. This was sort of my main point (sorry I did not express it more clearly). With web content, the author does not get to chose whether the document will be scrolled or paged or rendered aurally… The reader (via their choice of user agent and its settings) does. If Web Publications are to be web content, then the same is to be expected.

* You deleted "Note Authors or Publishers occasionally have strong preferences between scrolling and paginating". Why? This note seems useful to content authors and publishers who are used to different media that the web to understand why it is not possible to enforce "I want pagination".

* You deleted the note with a list of CSS specifications. I don't object to its removal, but it seemed like a useful addition for people who are not used to the web way of doing things, and whose first instinct when noticing a missing feature would be to propose a web-publication specific solution, instead of engaging with the right groups about the right specifications.

iherman:
Just a small note, @frivoal:

> * You deleted the note with a list of CSS specifications. I don't object to its removal, but it seemed like a useful addition for people who are not used to the web way of doing things, and whose first instinct when noticing a missing feature would be to propose a web-publication specific solution, instead of engaging with the right groups about the right specifications.

I have, just now, added a reference to the [roadmap that you have already seen](https://w3c.github.io/web-roadmaps/publishing/) to our WG home page, and [announced it to the WG](https://lists.w3.org/Archives/Public/public-publ-wg/2017Oct/0022.html). 

(B.t.w., I do not really agree with your "whose first instinct when noticing a missing feature would be to propose a web-publication specific solution" is entirely fair as far as this WG is concerned.)

iherman:
Merge per discussion on the call on 16. Oct. 2017

dauwhe:
> You deleted "Users must therefore rely on the UI features of the User Agent to navigate[...]" and the note that goes along with it. I understand the motivation is to not be overly prescriptive at this point of UA behavior, but I think this goes to far. I was not mandating any particular way the UA needs to make navigation possible, only that it does in some way. If we think that the UA shouldn't be in charge of that, and that this should instead be handled by the document/publication author in javascript, then we need to define some kind of API to make this doable. Otherwise, if there is no UA provided UI nor any API, the whole spec becomes mostly a no-op: while it would offer some meta-data conventions (assuming we do agree on a concrete manifest format), it would offer no no capabilities. I do not believe this was what we set out to do.

I agree that we need to say something about this; there's a section in the spec for UA behaviors, which I think is a better place. I'll see if I can move some of the text there.

dauwhe:
@frivoal wrote:

> The "Scrolling or Paginating" section has been largely rewritten. While I have no strong attachment to my original prose, which is of course perfectible, I don't particularly agree with the replacement:
> 

> >  Publications have historically been presented via paged media.

> Only if we take "publication" in a limited sense. Yes, books and other printed matter, as well as epub content have. If we consider everything on the web to also be a publication (which I think we should), then no. Posters, billboards, flyers, pamphlets, actual scrolls, and various other media types, ancient or modern are not paged either, and qualify just fine for being publications.

I didn't mean for the statement "Publications have historically been presented via paged media." to deny the existence or value of publications that use other methods. But our goal here is to motivate the discussion of pagination, and pointing out that most books, newspapers, magazines, journals, and newsletters over the last thousand years have been paginated seems reasonable. 

I think this section just needs to remind people that pagination is important, and is an issue for CSS.

dauwhe:
> You deleted the note with a list of CSS specifications. I don't object to its removal, but it seemed like a useful addition for people who are not used to the web way of doing things, and whose first instinct when noticing a missing feature would be to propose a web-publication specific solution, instead of engaging with the right groups about the right specifications.

I know a lot of ideas have been flying around in the early stages of WPUB, but I'm not really seeing strong evidence of people proposing web-publication-specific solutions for layout problems. Is there something you're particularly concerned about?

baldurbjarnason:
@dauwhe 
I can't speak for @frivoal but I know that a lot of people are worried about a recurrence of some of the very bad practices that the IDPF and epub reading systems were prone to. Given how dysfunctional the epub ecosystem has ended up—at least in part due to the habit of inventing and then badly implementing epub-specific solutions—it seems prudent to me to make it extra-clear in the spec itself that these tactics are not appropriate for web publications.

Personally I'd prefer to just have a note at the start that outlines common IDPF/epub reading system tactics that are unacceptable and must be avoided in a web publication context. Preferably written in a very angry tone, highlighted in red, surrounded by a blinking neon-coloured border and punctuated with swearing.

But since that'd almost certainly be considered 'hostile', I'm in favour instead of filling the spec with any sort of note, list, link, or reference that might help dissuade such practices.

frivoal:
@dauwhe 

> I didn't mean for the statement "Publications have historically been presented via paged media." to deny the existence or value of publications that use other methods. But our goal here is to motivate the discussion of pagination, and pointing out that most books, newspapers, magazines, journals, and newsletters over the last thousand years have been paginated seems reasonable.

> I think this section just needs to remind people that pagination is important, and is an issue for CSS.

I don't disagree with the essence of what you're saying here, but I am not sure how much value there is in reminding people who think that pagination is important (most of the publishing industry) that pagination is important. I'm ok with having that statement for others if you think it matters.

On the other hand, the second part, where you say that "Web pages almost always scroll", instead of web browsers, bothers me more. To me, the essential point to make (especially to the non-web publishing industry) is that the reader, not the medium, chooses the rendering and whether or not something is page. It is essential that people understand that while we can (and should) provide good tools to make pagination nice when the reader chooses to read things in the paginated mode of a UA that supports that, the decision whether to paginate or not is not up to the author/editor/publisher... Therefore, suggesting that paper/epubs paginate, while web pages scroll, is in my opinion counter productive.

frivoal:
@baldurbjarnason That's indeed the motivation for that note. I don't mind rewording or moving things around, but given the general level of concern and the historical issues, I think statements to that effect are appropriate, even if current participants have the best intentions in the world.

dauwhe:
> I don't disagree with the essence of what you're saying here, but I am not sure how much value there is in reminding people who think that pagination is important (most of the publishing industry) that pagination is important. I'm ok with having that statement for others if you think it matters.

I agree that the publishing industry thinks pagination is important. What we're trying to do is to convince the "web industry" that pagination is important. 

> On the other hand, the second part, where you say that "Web pages almost always scroll", instead of web browsers, bothers me more. To me, the essential point to make (especially to the non-web publishing industry) is that the reader, not the medium, chooses the rendering and whether or not something is page. It is essential that people understand that while we can (and should) provide good tools to make pagination nice when the reader chooses to read things in the paginated mode of a UA that supports that, the decision whether to paginate or not is not up to the author/editor/publisher... Therefore, suggesting that paper/epubs paginate, while web pages scroll, is in my opinion counter productive.

The vast majority of web content, as rendered by web browsers based on both author and user agent stylesheets, deals with vertical overflow by creating scroll bars. I mention this to point out that it's a major difference in user experience between the existing web and publications both printed and digital (as the majority of EPUB reading systems paginate by default).

In one sense the reader has a choice, in that they can choose to buy a paper book or magazine, or read an ebook, or read on the web. But most EPUB reading systems give more choice to the user  than browsers typically due, even to the point of offering a user preference for scrolling vs pagination. The rendering of web pages is greatly constrained by what browsers support, and so pagination is generally not an option unless you wisely hire Vivliostyle :)

But ebook reading system developers decided that pagination was often a better user experience for reading certain types of publications (like books), and went to great efforts to support that. I think this is good evidence that pagination is important for publications. 

Readers, authors, and publishers don't have choices at all unless the technology enables such choices. My goal, which I'm not sure is shared by everyone in the WG, is for web browsers to make providing a paginated experience easier, either with native support via CSS, or making the scripting of pagination easier with lower-level fundamentals like CSS Regions or Houdini custom layout.

TzviyaSiegman:
Let us attempt to put aside baggage about other specifications and systems. Reviewing this PR, I am content that it conveys our intentions to use existing CSS modules, avoid forking CSS, work with the CSSWG, and allow for user choice. We need to avoid inserting bias into specifications. We agreed to merge this PR at last week's [meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2017/2017-10-16-minutes.html#section6). So, I am going to merge it now. Please feel free to log issues if you think that there are issues left unaddressed.

PR #87, opened 2017-10-23
Navigation

mattgarrish:
Offering of prose to fill in empty holes in nav enhancement plus minor tweak to toc definition in infoset.


dauwhe:
Should we mention that HTML has a declarative syntax (rel=next) for indicating the next document in a sequence?

mattgarrish:
I added brief mention of the similarity. I think that's still a question that needs addressing. They're largely the same, but reading order provides the full sequence in one place (if fully specified) and doesn't limit sequencing to html. Links have had more traction in search than in UI, too.

iherman:
Discussed on 2017.10.23, merging

PR #93, opened 2017-11-07
Suggested an editors note in Section 4.5

prototypo:
Suggested an editors note in Section 4.5 to address concerns by Brady Duga, Benjamin Young, and David Wood raised by the resolution to Issue-76.

***
[Preview](https://s3.amazonaws.com/pr-preview/prototypo/wpub/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/fe59e27...prototypo:f53142d.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -648,7 +648,7 @@
 					</li>
 				</ul>
 
-				<p>A resource MAY link to one or more Web Publications to which it belongs.</p>
+				<p>A resource SHOULD link to one or more Web Publications to which it belongs.</p>
 				
 				<p class="ednote">A resource may link to multiple parent manifests. User agents may
                                    choose to present one or more alternatives to an end user, or choose a single
--- a/index.html
+++ b/index.html
@@ -134,7 +134,7 @@
 
 				<p>Web sites add another layer of relationship — this time between pages — but the relationship is a tenuous
 					one that typically depends on hyperlinks to add cohesion. Without a user that understands how to follow
-					the connections, a Web sit4.5 Linking to a Manifeste is still no more than a loose coupling of information.</p>
+					the connections, a Web site is still no more than a loose coupling of information.</p>
 
 				<p>The preceding is not a critique of the Web, but meant to highlight that the modern Web is very much an
 					active, event-driven experience. Users follow the necessary paths to obtain the information they


prototypo:
I have linked my GitHub and W3C accounts to resolve the contribution issues.

iherman:
I propose to merge it.

iherman:
One more thought: we should probably add an explicit key in the manifest for that purpose, to make it absolutely clear. Would it be worth adding this (as an editorial note for now) in the section on privacy?

@mattgarrish @prototypo

lrosenthol:
@iherman what do you want in the manifest?  This note is about something that goes in the resource, (optionally) connecting it back to the manifest

iherman:
We may need a term in the (JSON serialization of) manifest, something like:

```json
"privacy" : "<URL to the privacy statements>"
```

which is just to put in practice what the note refers to. It may not be important to have it in the FPWD _now_, but eventually we may need it, so just adding an editorial note that we will have that would not harm.

mattgarrish:
Did comments get mixed here, or am I missing something? What does privacy have to do with making it recommended to include a manifest link in each resource?

Is this comment made null by the other PR I added for the infoset, in other words? I didn't think we were going into actual serialization until of the infoset items until we tackle the manifest.

iherman:
Ok, forget my comment. It is too early to go there...

iherman:
@mattgarrish, will you merge this, or should we wait for more reviews? (I think it could be merged)

dauwhe:
I'm fine with merging this. But it does strike me as a feature that might end up at-risk, as implementation would not be fun, and I feel like the benefit is not so much for the reader as for the publisher.

lrosenthol:
I don't agree with that @dauwhe - I think it's an important reader feature as well to help you get back to the specific publication you started with (or perhaps even try out a different one)

mattgarrish:
What are we discussing? Is it:

1) that it is recommended to link to the manifest from all resources; or
2) that we should not allow linking to one or more resources?

This pull request only normatively addresses 1, so is there agreement that SHOULD is a good thing? If so, I think we can merge and then continue to address the second part.

The note strikes me as a potentially normative implementation detail that belongs in the lifecycle section. It's similar to what we adapted from WAM at one point. We might want to make it more note-y, by tweaking the beginning and adding a "future revision" clarification:

> When a resource links to multiple parent manifests, user agents may choose to present one or more alternatives to an end user, or choose a single alternative on its own. A user agent may choose to present whichever manifest it wishes based upon information that it possesses, even one that is not explicitly listed as a parent, e.g. based upon information it calculates or acquires out of band. In the absence of a preference by user agent implementors, selection of the first manifest listed is suggested as a default.
> These details will be formalized in a future version.

iherman:
> What are we discussing? Is it:
> 
>   1. that it is recommended to link to the manifest from all resources; or
>   2. that we should not allow linking to one or more resources?
>
> This pull request only normatively addresses 1, so is there agreement that SHOULD is a good thing? If so, I think we can merge and then continue to address the second part.

I do not think anybody expressed a wish for (2). Ie, I agree we can merge this...

>The note strikes me as a potentially normative implementation detail that belongs in the lifecycle section. It's similar to what we adapted from WAM at one point. We might want to make it more note-y, by tweaking the beginning and adding a "future revision" clarification:
>
> ...

Do you propose the current proposed _Editor's Note_ for a (real) _Note_ with your text? If so, I see your text as a more spec-y text that can indeed replace the current editor's note.

mattgarrish:
> I do not think anybody expressed a wish for (2).

Okay, I thought it was the "or more" part that was causing this disagreement.

> Do you propose the current proposed Editor's Note for a (real) Note with your text?

Yes, I find the note almost normative sounding, so with a couple of tweaks I think we can lose the "editor's" designation and leave it as is or merge it somewhere else later.

mattgarrish:
I'm going to merge this one and I'll do the minor tweaking to the note later.

PR #96, opened 2017-11-09
add relationship to wam section

mattgarrish:
In looking at adding a note to reflect the discussions at TPAC, I found that both the infoset and manifest are likely influenced by any future harmonization. Rather than pick one section for a note, I've added to a new "relationships" section in the introduction.

Comments on this approach, and the substance of what is stated welcome.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/webappmanifest-note.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/fe59e27...30ebe0d.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -203,6 +203,47 @@
 					work that can be represented on the Web constitutes a potential Web Publication.</p>
 			</section>
 
+			<section id="relationships">
+				<h3>Relationship to Other Specifications</h3>
+
+				<section id="rel-wam">
+					<h4>Web App Manifest</h4>
+
+					<div class="ednote">
+						<p>The working group is investigating integration with the work done on [[appmanifest]] to avoid
+							duplication of concerns both in the expression and serialization of a manifest and in the
+							processing and compilation of the infoset. It is expected, however, that there will be
+							differences in the processing models of each (e.g., the initiation of the reading experience for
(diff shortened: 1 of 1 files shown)


iherman:
@mattgarrish, 

Thanks for starting this! Comments:

> The initial stage of evluating this assumption involves:

evluating -> evaluating

>
> The next stage will involve discussions about how the expression of manifests can be aligned (e.g., has Web Publications defined members that are generally useful for applications), and how to separate the different concerns and requirements of apps and manifests so that unwanted behaviors are not inherited.

I think we can say a bit more (based on our discussion). Something like:

"The next stage will involve the serialization of the infoset in JSON, and see how this expression of the manifest can be aligned with the expression of the [[appmanifest]]. Wherever semantically possible, the terms used and defined for [[appmanifest]] should be reused, and, conversely, some of the members defined for Web Publications may be generally useful for applications and may be added to the [[appmanifest]] as a common pool of information. To avoid possible future conflicts, care should be taken to use a naming schemes that clearly separates terms to be used for Web Publications only."

iherman:
Thanks for the changes. Ship it!!

lrosenthol:
Agreed with @iherman - merge it!   (thanks @mattgarrish)

PR #97, opened 2017-11-09
clarify in scope the intent to facilitate different UA models

mattgarrish:
Not asked for, but this seems to be a question we realize we're all struggling to resolve in our minds as we consider how initiation and consumption of web publications occurs.

I think it would be good to say something about staying open to different models, but won't shed any tears if it passes a short unloved life and disappears.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/target-ua.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/fe59e27...38bdbad.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -201,6 +201,15 @@
 
 				<p>Further, this specification does not attempt to constrain the nature of a Web Publication: any type of
 					work that can be represented on the Web constitutes a potential Web Publication.</p>
+
+				<p>The specification is also intended to facilitate different user agent architectures for the consumption of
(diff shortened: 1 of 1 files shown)


dauwhe:
I think this should have a short life as a pull request, and a long life as an actual part of the document. I vote merge!

iherman:
Merge it before the different PR begin to creating chaos:-)

lrosenthol:
I agree - merge it!

mattgarrish:
Will do. Just wanted to write them all up before memories faded.

Issue #94, opened 2017-11-08
What does a WPub URL resolve to?


prototypo:
# Problem Statement

We have decided that a WPub with be identified with a URL. What should that URL resolve to?

This issue suggests four scenarios and evaluates their ramifications in the hopes of driving toward consensus. A conclusion section at the bottom of this post suggests an answer to the question posed.

# Scenarios

The working group has considered several possible answers to this question. They mostly fall into one of four scenarios:

1. A WPub URL resolves to a JSON manifest file
2. A WPub URL resolves to an HTML file that contains both a table of contents (TOC) and other metadata
3. A WPub URL resolves to an HTML file containing a TOC and which links to a JSON manifest file
4. A WPub URL resolves to a binary package (e.g. a ZIP file or SQLite database file)

Each of these scenarios are considered in turn as they would be viewed by four types of clients:

- An existing search engine bot
- A WPub aware search engine bot
- A reader using an existing user agent
- A reader using a WPub aware user agent

## Scenario 1: JSON Manifest

If a WPub URL resolves to a JSON manifest file, the four clients may be expected to act like this:

- An existing search engine bot (path 1A in the image below) will not index JSON files.
- A WPub aware search engine bot (path 1B in the image below) will need to index all JSON files on the Web to determine which ones represent WPubs.
- A reader using an existing user agent (path 1C in the image below) will be presented with a raw JSON representation.
- A reader using a WPub aware user agent (path 1D in the image below) will be able to see and operate upon the WPub as intended.

Scenario 1 seems inconvenient for both old and new clients with the exception of new WPub aware user agents.

![wpub-canonicalresourcejson](https://user-images.githubusercontent.com/824236/32531863-3599fdee-c3fb-11e7-8d93-e2261c34278c.png)

## Scenario 2: HTML TOC & Metadata

If a WPub URL resolves to an HTML file that contains both a table of contents (TOC) and other metadata, the four clients may be expected to act like this:

- An existing search engine bot (path 2A in the image below) will encounter and index the HTML page without modification.
- A WPub aware search engine bot (path 2B in the image below) will be able to determine that the HTML page represents a WPub, but will need to parse the entire file in a WPub-specified manner to extract metadata.
- A reader using an existing user agent (path 2C in the image below) will see a TOC containing links to the components of the WPub.
- A reader using a WPub aware user agent (path 2D in the image below) will be able to see and operate upon the WPub as intended.

[...]

PR #98, opened 2017-11-10
Added a paragraph on the landing page in the lifecycle section

iherman:
This is an attempt to include in the document the emerging consensus on issue #94.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/landing-page.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/256a2e9...072e663.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -776,9 +776,13 @@
 		<section id="wp-lifecycle">
 			<h2>Web Publication Lifecycle</h2>
 
+			<p>As described in <a href="#wp-address"></a>, a Web Publication has an <a>address</a> that can be dereferenced. That URL MUST resolve to an HTML file (referred to as a <dfn>landing page</dfn>), and this HTML file MUST include a link to the <a>manifest</a> as described in <a href="#manifest-linking"></a>. The <a>landing page</a> SHOULD also include an HTML serialization of the <a href="#enh-toc">Table of Content</a>.</p>
+
+			<p class=ednote>The exact definition of how the TOC should be added to the landing page is still missing; most probably it would be a <code>nav</code> element with some predefined identifier and/or class.</p>
+
 			<div class="ednote">
 				<p>The publishing working group is currently evaluating the best approach for implementing web publications
-					in user agents. This note is intnded to provide an overview of where current thinking is and what issues
+					in user agents. This note is intended to provide an overview of where current thinking is and what issues
 					are under consideration.</p>
 
 				<p>The development of web publications is not viewed as a separate forking of the web, but an enhancement
--- a/index.html
+++ b/index.html
@@ -583,6 +583,8 @@
 					<li>calculate a table of contents using its own algorithms.</li>
 				</ol>
 
+				<p class=ednote>This section should be brought in sync with the section on <a>landing page</a>.</p>
+
 				<p class="issue" data-number="9">The question is whether the manifest/infoset needs to provide navigation, or
 					whether such mechanisms belong in the content.</p>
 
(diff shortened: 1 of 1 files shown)


mattgarrish:
I only wonder if these requirements belong under the Address section itself (better understanding by proximity).

Substantively, though, I have no problem with it. We can always tweak locations later.

lrosenthol:
Why does the landing page have to be HTML?  Why can't it be SVG or an image?

dauwhe:
> @iherman Why does the landing page have to be HTML? Why can't it be SVG or an image?

An image can't contain a link to the manifest.

lrosenthol:
@dauwhe it can in the link header, which is one of the ways to link back (as described elsewhere in the doc)

mattgarrish:
HTML may be restrictive, but the point made by Florian that images and audio and video are not the "web" was a good one and still resonates with me. Even if there are ways of rendering such media for WPUB-aware user agents, they can't be made usable for vanilla implementations. A link header is effectively useless. Asking for one HTML file for those media types seems like a fair trade-off for usability.

I personally feel like we're biting off more than we can chew in taking such a liberal view that anything that can be hosted on a web server is a potential web publication or candidate for inclusion in the default reading order, but I don't want to cause a 100 comment argument here... :)

lrosenthol:
I didn't hear Florian say that, but I disagree completely.  The OWP is well defined (<https://www.w3.org/wiki/Open_Web_Platform>) and clearly includes more than HTML.

mattgarrish:
They're are technologies used on the Web, yes, but they are not how users navigate the Web, which is the point here. HTML is the backbone of the web that brings those resources together.

I can't move from jpeg to jpeg, for example, or audio clip to audio clip. We're attempting to solve that here, but in a way that isn't compatible with the Web as it exists.

iherman:
Coming back to the original comment of @lrosenthol: yes, indeed, we do refer to the possibility of referring to the manifest via the HTTP return header. However:

* It turns out to be very complicated for lambda users (and many system managers) to set up such return headers correctly
* Although I am not familiar with all the details, I believe that browsers would not make it easy for extensions, javascript programs, whatever, to have access to the return header of an HTML file (it is not part of the DOM) let alone the return header for an image

Let me add: alas! The HTTP return headers is an underutilized part of the architecture (I think), but we must be pragmatic. The proposal referring to the `<link>` element tries to be pragmatic and avoid shooting in our own foot...

lrosenthol:
Well, we at least agree on SVG - so MUST be HTML won't work.

iherman:
@lrosenthol the "real" content may be SVG (e.g., a series of Manga images). That is not contradictory to a requirement that the landing page is in HTML.

SVG does not have a `<link>` element that we could use, for example.

BigBlueHat:
@iherman this PR mentions not having a definition for the structure of the Table of Contents, but farther down the document we do have that definition: https://w3c.github.io/wpub/#wp-table-of-contents

Other than that, I'm +1.

iherman:
Oops @BigBlueHat, good catch! I have put a reference to that section, removed the editor's note and added an editor's note in the section you referred to that it should be brought to sync with this new addition. I am too tired to do anything else right now (and I travel back home tomorrow).

I may try to propose a sync in [that section](https://w3c.github.io/wpub/#wp-table-of-contents) on Monday, unless @mattgarrish beats me into it:-)

mattgarrish:
What I'm going to do here is merge this prose but suggest we leave open #94 for further discussion (i.e., in the vein of just because we've added prose doesn't mean it is set in stone).

In addition to looking at toc overlap, I'd like to do some further restructuring, such as setting up a "content" section for the landing page description and also the "linking to a manifest" section. I'd also like to do some more general review before fpwd, and leaving this open gets in the way.

HadrienGardeur:
In reply to @iherman 

> SVG does not have a `<link>` element that we could use, for example.

No need for such an element when you can rely on the `Link` header in HTTP.

iherman:
@HadrienGardeur this is of course related to the #94. You are right that, in general, a Link header would suffice, but this comment came about on whether a pure SVG file could play the role of a landing page or not.

PR #99, opened 2017-11-12
add construction section

mattgarrish:
This is the thought I had after seeing the landing page section. The landing page is more of a content construction section than anything to do with the lifecycle. The section on linking to a manifest was similarly out of place, as it is about a content requirement not the manifest itself.

The one thing new I added was a resources section that says that there are no restrictions on the media types of web publication resources. We've been working under that assumption, but for understanding outside the working group I think we need to be more explicit about this point.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/construction-section.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/7806641...feea193.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -278,8 +278,8 @@
 					<dt><dfn data-lt="Manifests">Manifest</dfn></dt>
 					<dd>
 						<p>A manifest represents structured information about a <a>Web Publication</a>, such as informative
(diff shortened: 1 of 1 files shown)


iherman:
LGTM

mattgarrish:
Merging in the interests of moving on to other cleanup, but if anyone spots anything please just open a new issue in the tracker.

PR #101, opened 2017-11-14
update landing page to match issue discussion in #94

mattgarrish:
This is just an attempt to capture where #94 has gone. It may still be too strict in some spots, and I'll update as I see the discussions shift.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/landing-page.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/f447f71...b7e961e.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -815,20 +815,17 @@
 				<h3>Landing Page</h3>
 
 				<p>As described in <a href="#wp-address"></a>, a <a>Web Publication</a> has an <a>address</a> that can be
-					dereferenced. The address URL MUST resolve to the <dfn>landing page</dfn>, an [[!html]] document that
+					dereferenced. This URL MUST resolve to the <dfn>landing page</dfn>, which is an [[!html]] document that
 					represents the primary entry point for the Web Publication.</p>
 
-				<p>Unlike other resources, the landing page MUST <a href="#wp-linking">link to the manifest</a> to ensure
-					that user agents can discover the Web Publication.</p>
+				<p>The landing page MUST be a resource of the Web Publication, but its inclusion in the <a>default reading
+						order</a> is OPTIONAL. Unlike other resources, the landing page MUST <a href="#wp-linking">link to
+						the manifest</a> to ensure that user agents can discover the Web Publication.</p>
 
-				<p>To ensure that the Web Publication is consumable in different types of user agents, including those that
-					do not support Web Publications, the landing page SHOULD include an [[!html]] serialization of the <a
-						href="#enh-toc">table of contents</a> (see <a href="#wp-table-of-contents"></a> for more
-					information).</p>
-
-				<p class="issue" data-number="94">Additional questions about whether specifying HTML is necessary, whether
-					the resource has to be part of the web publication or not, and whether the table of contents is necessary
-					to specify at this location if provided in the content.</p>
+				<p class="note">To improve the usability of the Web Publication, particularly in user agents that do not
+					support Web Publications, provide access on the landing page to navigation aids that facilitate
+					consumption of the content (e.g., include a <a href="#enh-toc">table of contents</a> on the page, or
+					provide a link to one).</p>
 			</section>
 		</section>
 		<section id="wp-lifecycle">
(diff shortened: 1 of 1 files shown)


mattgarrish:
Yes, I'm not saying this is a consensus proposal at this point, or ready for inclusion. I just wanted to start updating the text so there's something to read.

BigBlueHat:
@mattgarrish I think you handled the situation nicely and am 👍 on merging this and moving forward.

index.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+la<!DOCTYPE html>
```
BigBlueHat: This is the one thing I don't approve of in this text. 😉

HadrienGardeur:
👎  from me as long as we keep the following requirements:

- The entry page MUST be a resource of the Web Publication (https://github.com/w3c/wpub/pull/101/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR819)
- the entry page MUST link to the manifest (https://github.com/w3c/wpub/pull/101/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR820)

These are IMO TBD and it's not clear if these requirements can be realistically achieved.

index.html
mattgarrish: Lol, I would have caught it eventually...

BigBlueHat:
> These are IMO TBD and it's not clear if these requirements can be realistically achieved.

@HadrienGardeur you're going to have to be clearer what you mean by "realistically achieved" and what things prevent those from being MUSTs.

mattgarrish:
Given that the thread was about adding these, I can't see taking them out as reasonable to closing #94. We don't have complete agreement, obviously, but the numbers are on the side of this interpretation right now.

Let's pick up their merits in more focused issues.

HadrienGardeur:
There's no consensus on the entry page being a resource of the Web Publication, several other members of this group have also pointed that out, including @GarthConboy in https://github.com/w3c/wpub/issues/94#issuecomment-344137566

BigBlueHat:
@mattgarrish this line could use some clarification--and may be part of the conflation between identification and discovery:

> Unlike other resources, the entry page MUST link to the manifest to ensure that user agents can discover the Web Publication.

Do you mean "discover that this thing is a Web Publication?" or do you mean something more vague like "search engine discovery of Web Publications?"

If you're referring to "identification *as* a Web Publication," then that text needs revision.

BigBlueHat:
@HadrienGardeur that was also in the context of "landing page" being conflated with an advertising page--which is not the intent here. The goal of this (and @mattgarrish's wise renaming to "entry page") is to have a known and predetermined extensible format sent in response to a request for a Web Publication.

There's still lots to be determined about what role(s) this entry page might play in the spec and in implementations.

HadrienGardeur:
@BigBlueHat sorry but I don't see your point. Renaming "landing page" to "entry point" is something completely different from saying that the "entry point" MUST be within the boundaries of the WP.

BigBlueHat:
@HadrienGardeur well, at this point, filing an issue about the MUST is the best course of action as these comments are all quite hidden post-merge.

HadrienGardeur:
How convenient, let's merge things first then delay further discussions until the end of time...

BigBlueHat:
@HadrienGardeur not delaying further discussion. Just trying to focus discussion. File the issue, and lets talk there. If you feel it really is broken this way, then it's worthy of its own problem/solution definition and focused discussion.

GarthConboy:
And, FWIW, I'm fine with merging and starting a new issue, but in response to @HadrienGardeur:

- The entry page MUST be a resource of the Web Publication 
- The entry page MUST link to the manifest

I think the first is, indeed, TBD.  The second, however, I think should be retained.

mattgarrish:
> If you're referring to "identification as a Web Publication," then that text needs revision.

Sorry, I saw this and then lost track of it. Yes, it's the former. Maybe this is all that needs saying:

> the entry page MUST link to the manifest to ensure that user agents can locate that document.

mattgarrish:
An additional change was made in PR #105 to change the requirement the entry page be a publication resource from MUST to SHOULD.

The requirements are:

MUST be an HTML document
MUST include a link to the manifest
SHOULD be a publication resource

Issue #103, opened 2017-11-15
Role of the HTML document returned by the WPUB URL


HadrienGardeur:
We've had long discussions in issue #94 about what the WPUB URL resolves to and there's a consensus that it should return an HTML document.

What's not well defined though is the exact role of that document.

Several proposals have been made so far:

- discover the Web Publication (through a link to its manifest)
- provide information about the Web Publication
- enable the acquisition of a Web Publication (if the publication requires authentication or is behind some sort of paywall)
- provide a Web App or a polyfill to handle various WP specific features (suggested by @BigBlueHat in this thread)

There's also an on-going discussion whether this document MUST belong to the Web Publication itself or remains external to the publication (I'll let @mattgarrish and @GarthConboy repost some of their relevant comments here).
Every agrees that the document MAY belong to the publication.

If the document belongs to the Web Publication, it MAY be:

- the table of contents
- the cover (not a concept that's been clearly discussed or defined at this point in time)
- the recommended resource to start reading the publication (could be different from the first resource in the reading order)

dauwhe:
Hadrien, could you describe how it would work if the HTML document returned by the WP URL is not part of the WP, and doesn't contain a link to the WP's manifest?

baldurbjarnason:
> Hadrien, could you describe how it would work if the HTML document returned by the WP URL is not part of the WP, and doesn't contain a link to the WP's manifest?

Is there any reason why it wouldn't contain a link to the manifest? E.g. for RSS, sub-pages (like 'about') are generally not a part of what appears in the feed but still contain the link. Was there a specific reason cited in that other thread of why that can't be the case?

I have to confess that issue https://github.com/w3c/wpub/issues/94 totally lost me about halfway through so I may well have missed this point if it was made later on.

HadrienGardeur:
@dauwhe that's a little more specific than the question that I just raised in this issue.

First of all, I think that the document will always return information about the publication, so out of the three things that I've listed, this one is always true.

I'll use concrete examples:

**Info about the Web Publication, no link to the manifest, not part of the WP**

This could apply to a book behind a paywall, for example Hachette Book Group could start selling Web Publications directly from their website. 

If the user is not logged in and/or hasn't bought that specific publication, he/she gets redirected to a product page with info about the book (use case 1) and a way to buy it (use case 3).

**Info about the Web Publication, with a link to the manifest, not part of the WP**

This could apply to a publisher distributing strictly Open Access content.

[...]

PR #105, opened 2017-11-17
updates per comments in issue #103

mattgarrish:
I'm not sure if this capture everything given the lengthy discussion. Substantively, only the web publication resource requirement appears to change from MUST to SHOULD. I have made a couple of tweaks to the toc and default reading order to make it a little more clear that these come from publication resources.

I'm not clear why we would start talking about scripting in the entry page, so avoided that. If we want to describe interfaces, shouldn't it be a separate section with a recommendation to use the entry page if the interface can only be initiated from one page?

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/address.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/4c1aaf0...a041bbb.html)

### DIFF
--- a/index.html
+++ b/index.html
@@ -607,8 +607,10 @@
 
 					<p>The default reading order MUST include at least one resource.</p>
 
-					<p>The default reading order is either specified directly in the manifest or a link is provided to an
-						[[!html]] <code>nav</code> element whose list of links are processed to create one.</p>
+					<p>The default reading order is either specified directly in the manifest or in an [[!html]]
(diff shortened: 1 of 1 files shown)


mattgarrish:
Given the lack of comment, I'm going to merge this on the premise it meets the "livable with" criteria for now.

mattgarrish:
Reiterating here, current proposal is:

MUST be an HTML document
MUST include a link to the manifest
SHOULD be a publication resource

PR #107, opened 2017-11-19
expand on resource list definition

mattgarrish:
Given the importance of establishing the bounds of the publication, and that it is more important that the resource list should be complete than that it might not always be, this is a rework to improve the resource list definition. Improvements always welcome.

***
[Preview](https://s3.amazonaws.com/pr-preview/w3c/wpub/resource-list.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/cb16a5c...027828a.html)


iherman:
Let us ship it...

PR #109, opened 2017-11-21
Add manifest example to spec, and add sample document to repo

dauwhe:
Even though we have made no decisions on the serialization of the manifest, I think the spec will be easier to understand for newcomers if we had a small example of what a WP manifest might look like. An example is worth a thousand words! So I've added a very simple one, with a note saying this is only illustrative. I expect the final serialization will be more complicated than this. 

In the same spirit, I've also made an actual sample web publication.

***
[Preview](https://s3.amazonaws.com/pr-preview/dauwhe/wpub/master.html) | [Diff](https://s3.amazonaws.com/pr-preview/w3c/wpub/b015eae...dauwhe:16ea3fe.html)


lrosenthol:
I think an example that uses any sort of serialization that we have not agreed on will cause more confusion...

iherman:
> I think an example that uses any sort of serialization that we have not agreed on will cause more confusion...

In general, I would agree with this, @lrosenthol. But, in this case, there is only a minimal example (duly noted as temporary) that just sets the scene. We do have a decision that the manifest will be serialized in JSON, actually. Ie, having that example may not be harmful and gives a direction...

mattgarrish:
Would it make sense to create a samples repository and close this PR? I suspect, like with the epub samples, we'd want to open access to more people, so having them in the main repo is a bit concerning. It would also give us more flexibility to potentially publish samples as github pages so we could have live demos.

Until we solve the serialization, even simple attempts at manifest examples within the specification are going to prove contentious.

iherman:
Works for me

Issue #124, opened 2018-01-24
Expressing metadata in multiple scripts/languages


HadrienGardeur:
The current WP infoset is fairly consistent with the WAM:

- there's a single language element meant to indicate the language of all properties expressed in the manifest
- same thing for the direction

This is quite different from EPUB 3.x where each metadata can be expressed in multiple scripts/languages. Here's an example from the 3.1 spec:

```
<dc:creator opf:alt-rep-lang="ja" opf:alt-rep="村上 春樹">
    Haruki Murakami
</dc:creator>
```

Readium and the RWPM also provide support for multiple scripts/languages per property:
```
"author": {
  "name": {
    "ru": "Михаил Афанасьевич Булгаков",
    "en": "Mikhail Bulgakov",
    "fr": "Mikhaïl Boulgakov"
  }
}
```

Since the Japanese publishing industry (ping @frivoal) told us multiple times in the past that this is very important for them, I'm wondering if the current direction for WP is on purpose or not.

TzviyaSiegman:
@murata2makoto please comment on behalf of JEPA

murata2makoto:
Expressing each piece of metadata (e.g., titles and author names) in multiple scripts (CJK ideographics and some phonetic script (e.g, Kana)) is a must for Japanese.  This is because there are multiple and equally reasonable phonetics for some names.  For example, 智子 might be pronounced as either Tomoko or Satoko.  We really do not know.

HadrienGardeur:
Thank you @murata2makoto for this feedback.

I think that we need to carefully reconsider what's in our infoset. Here are a few notes:

- the requirements for Japanese are incompatible with the design of the WAM as well
- RWPM allows more than one "alternative representation" which should be a perfect fit (see example above)

I'll also open a separate issue about reading direction, this is a related problem but it's more difficult to solve IMO.

iherman:
My approach would be:

- It is o.k. to have the current language+direction setting as a _default_ for the metadata. This can cover many of the use cases at least in US and in Europe.

- (This is still hazy) everywhere where we would have an essentially textual value for the information item, we should allow to have, instead, a structure consisting of a value, language tag and direction. This is fine in simple JSON, I still have to see how this can properly be done (maybe with some JSON-LD 1.1 features) in JSON-LD.

I think we should include the consideration on directions into this issue right from the start.

HadrienGardeur:
RWPM has a slightly different approach:

[...]

PR #129, opened 2018-02-02
Change in the language/direction infoset items

iherman:
Added/changed the language and direction section to allow for each text to be localizable, which led to a slight reorganization of the spec.

Refers to #124

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/129.html" title="Last updated on Feb 8, 2018, 5:56 AM GMT (adf8ad9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/129/1955a17...adf8ad9.html" title="Last updated on Feb 8, 2018, 5:56 AM GMT (adf8ad9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -402,10 +402,10 @@
 							<span>RECOMMENDED:</span>
 							<ul class="flat">
 								<li><a href="#wp-a11y">accessibility</a></li>
-								<li><a href="#wp-language">base direction</a></li>
+								<li><a href="#wp-language-and-dir">base direction</a></li>
 								<li><a href="#wp-canonical-identifier">canonical identifier</a></li>
 								<li><a href="#wp-creators">creator</a></li>
-								<li><a href="#wp-language">language</a></li>
+								<li><a href="#wp-language-and-dir">language</a></li>
 								<li><a href="#wp-mod-date">modification date</a></li>
 								<li><a href="#wp-privacy">privacy policy</a></li>
 								<li><a href="#wp-pub-date">publication date</a></li>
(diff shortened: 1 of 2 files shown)


iherman:
(The travis error can be ignored, it was the result of an unlucky manipulation with .travis.yml)

index.html
```diff
@@ -552,30 +528,50 @@ <h4>Creators</h4>
...
 					<ul>
 						<li>use the <a>non-empty</a> language declaration of the manifest;</li>
-						<li>use the first non-empty language declaration found in a resource in the <a>default reading
-								order</a>;</li>
+						<li>use the first non-empty language declaration found in a resource in the <a>default reading order</a>;</li>
 						<li>calculate the language using its own algorithm.</li>
 					</ul>
 
-					<p>If a language tag cannot be determined, the value "<code>und</code>" (undetermined) MUST be used.</p>
+					<p>If a language tag cannot be determined, the value "<code>und</code>" (undetermined) MUST be used. If the base direction cannot be determined, the value <code>ltr</code> MUST be assumed.</p>
```
HadrienGardeur: `auto` might be a better default.

iherman: oops... of course. Will take care of this.

index.html
```diff
@@ -403,7 +402,7 @@ <h3>Requirements</h3>
 							<span>RECOMMENDED:</span>
 							<ul class="flat">
 								<li><a href="#wp-a11y">accessibility</a></li>
-								<li><a href="#wp-dir">base direction</a></li>
+								<li><a href="#wp-language">base direction</a></li>
```
lrosenthol: Why is this changed to language, it should still be dir?

iherman: @lrosenthol the previous version contained two independent sections, one for language and another for direction. With the extra features it looked editorially better to merge the two sections. Hence the two identical targets. 

The "id" value may not be ideal (just kept the old id of one of the original sections) but that is only an internal thing, invisible to the reader. We can harmonize those later.

Issue #126, opened 2018-01-31
Progression/direction between resources


HadrienGardeur:
The current infoset does not contain any information about the reading progression (progression between resources) for a given publication.

In EPUB, this was handled using an attribute on the `spine` (the EPUB equivalent for the default reading order), and it's handled through `direction` in the `metadata` section of the RWPM.

This is a very important information for User Agents since it's absolutely necessary for providing key affordances.

For instance, in the case of Readium-2, we use three webviews side by side on iOS and Android:
```
_____________________________________
|           |           |           |
| Webview 1 | Webview 2 | Webview 3 |
|           |           |           |
```

If the reading progression is set to `ltr`, then:

- 1 contains the previous resource (or none if we're displaying the first one in the reading order)
- 2 contains the current resource
- 3 contains the next resource (or none if we're displaying the last one in the reading order)

If the reading progression is set to `rtl`, then:

- 1 contains the next resource (or none if we're displaying the last one in the reading order)
- 2 contains the current resource
- 3 contains the previous resource (or none if we're displaying the first one in the reading order)

In addition to setting up these Webviews, this reading progression is also used for:

- mapping tap zones that will allow us to get to the next page (if a resource is paginated) along with the next resource
- specific touch gestures, such as swipe/drag
- keyboard events, such as mapping arrow keys to move forward/backward in the publication

HadrienGardeur:
@laudrain this is the issue that I mentioned during our call today. Could you post additional use cases and example as well?

dauwhe:
How would this infoset metadata relate to the [page progression](https://drafts.csswg.org/css-page-3/#progression) direction of the primary HTML resources of the publication? What if the infoset value contradicts the value in some or all of the primary resources?

HadrienGardeur:
Page progression is about the **reading direction** of **a single resource**.

This is about the progression **between resources**.

When each resource of a publication is scrolled, the reading direction can't contradict this value.

Paged media is different though. 

Setting `ltr` on the publication when all resources use `rtl` would not really "contradict" anything, but you might end up with a weird user experience where you get to the previous resource instead of the next one if the UA allows to swipe/tap inside paginated resources as well as between resources.

That said, this inconsistency goes beyond that. For example, what happens if the first chapter uses `ltr` and the second chapter uses `rtl` as its page progression ?

iherman:
[...]

PR #142, opened 2018-02-19
Added a new information item for progression direction.

iherman:
Fixes issue #126

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/142.html" title="Last updated on Mar 1, 2018, 9:32 AM GMT (679f6d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/142/5319fb5...679f6d1.html" title="Last updated on Mar 1, 2018, 9:32 AM GMT (679f6d1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -421,8 +421,14 @@
 					</dd>
 					<dt>Structural Properties</dt>
 					<dd>
-						<p>REQUIRED: <a href="#wp-default-reading-order">default reading order</a> and <a
-								href="#wp-resource-list">resource list</a></p>
+						<div>
+							<span>REQUIRED:</span>
+							<ul class="flat">
+								<li><a href="#wp-default-reading-order">default reading order</a></li>
+								<li><a href="#wp-resource-list">resource list</a></li>
+								<li><a href="#wp-reading-progression">reading progression</a></li>
+							</ul>
+						</div>
 						<p>RECOMMENDED: <a href="#wp-table-of-contents">table of contents</a></p>
 					</dd>
 				</dl>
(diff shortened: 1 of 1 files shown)


dauwhe:
As @BigBlueHat mentioned on the call, there are only two values for page progression direction. Within HTML resources, CSS does allow for overflow in both x and y axes. If a publication resource scrolls "down," it might be a better reading experience for the next resource to appear below. Do we need a "top to bottom" direction?

iherman:
@dauwhe, better discuss this on #126, because this PR is really only a write-up of the discussion and conclusion that happened there...

(I will copy this issue over there, just to save you time...)

gardenia-corp:
How about default direction to "Reading System can choose default direction" which is same as idpf defined (http://www.idpf.org/epub/30/spec/epub30-publications.html#sec-spine-elem)? It mightbe convenient when we need to convert from idpf to web infoset.

rakutenjeff:
How about default direction to "Reading System can choose default direction" which is same as idpf defined (http://www.idpf.org/epub/30/spec/epub30-publications.html#sec-spine-elem)? It mightbe convenient when we need to convert from idpf to web infoset.

iherman:
@rakutenjeff : I am fine with what you propose and happy to do it if we have a consensus on the PR overall...

mattgarrish:
My take is we merge this PR and continue to work out any additional details in separate issues, so that's what I'll do. :)

Nothing is written in stone yet.

rakutenjeff:
Thanks @iherman Do you mind if I record it as a new issue since this is merged? Beside the default progression I was hoping wpub could have consistency with epub. And further more, my image of relationship between wpub and epub is wpub define what should be in manifest (and maybe define json serialization as well), and epub address those manifest items in packaged wp (epub) domain. So the manifest should be compatible. The benefit is current publisher who already has millions of epubs would be easy to step into WP stage in the future.

iherman:
@rakutenjeff of course, by all means! Please do.

PR #158, opened 2018-03-03
ORCID logo rendering done properly

iherman:
Added a separate postprocessing script to respec to extract the ORCID ID from the initial config and display the ORCID logo. The ORCID Id should now be a separate key in the persons’ structure.

@mattgarrish I tried to remove the underlining of the logo by adding an extra entry in the CSS file. Did not work:-( Maybe you will be luckier!

If it is o.k. with you, I will make the changes on the annotation document as well

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/158.html" title="Last updated on Mar 3, 2018, 1:30 PM GMT (e32a830)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/158/3e1d32f...e32a830.html" title="Last updated on Mar 3, 2018, 1:30 PM GMT (e32a830)">Diff</a>

### DIFF
--- a/common/css/common.css
+++ b/common/css/common.css
@@ -28,5 +28,5 @@ dfn {
 }
 
 span.orcid a {
-	text-decoration: none !important;
+	border-bottom: none !important;
 }
--- a/common/css/common.css
+++ b/common/css/common.css
@@ -26,3 +26,7 @@ dt, dd {
 dfn {
 	font-weight: bold !important;
 }
+
+span.orcid a {
+	text-decoration: none !important;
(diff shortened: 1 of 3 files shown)


mattgarrish:
It wasn't text-decoration but a border-bottom setting that was putting the underline under the images. I pushed a fix that removes. There's still a bit of the background change apparent when you hover, but we should probably leave at least one visual clue.

I've patched the wpub-ann file, too.

PR #171, opened 2018-03-22
update accessibility section in infoset to be a link to a report

mattgarrish:
Here's a thought on how we can address the current accessibility section. This PR would turn it into a link to a report.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/171.html" title="Last updated on Mar 22, 2018, 7:17 PM GMT (2581eed)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/171/1af873d...2581eed.html" title="Last updated on Mar 22, 2018, 7:17 PM GMT (2581eed)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -399,7 +399,7 @@
 						<div>
 							<span>RECOMMENDED:</span>
 							<ul class="flat">
-								<li><a href="#wp-a11y">accessibility</a></li>
+								<li><a href="#wp-a11y">accessibility report</a></li>
 								<li><a href="#wp-language-and-dir">base direction</a></li>
 								<li><a href="#wp-canonical-identifier">canonical identifier</a></li>
 								<li><a href="#wp-cover">cover</a></li>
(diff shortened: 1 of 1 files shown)


laudrain:
it seems to me that there may be confusion between the different reports cited in this section.
For instance, which report should be included as a resource in the WP: the reports of the first paragraph or the one in HTML format of the third paragraph?

avneeshsingh:
- In accessibility standards groups (in DAISY as well as W3C), the concept of conformance/accessibility report came up as a major feature so it is good to have it in WP/PWP. At the same time there is parallel discussion going on for support for external metadata. How will both of these work together?
- We also need to think how external records like ONIX will be managed because ONIX also have good amount of accessibility metadata. If we end up providing separate facility for external records like ONIX, then will we duplicate ONIX accessibility codes in this report? 

Or one way can be to keep accessibility metadata with other ONIX/schema.org metadata and provide conformance information and description of schema.org properties and ONIX codes in this accessibility report.

mattgarrish:
I think what you're capturing @avneeshsingh is a further discussion about whether the metadata belongs in a consolidated format somewhere, or whether it is better to have one accessibility report.

This PR is only trying to make sense of why we have something called "accessibility" in the infoset. Short of adding every schema.org property to the infoset, which I don't think is realistic or useful, if we are going to retain a property making it a link to a report seems the most useful approach.

I'm fine adding a further note that this property may be removed and/or machine-processable accessibility metadata may also be included elsewhere. I didn't want to make the machine-processable metadata a requirement of this report for these reasons.

@laudrain I'm not sure what you mean by separate reports? There is only one report link. It is recommended the report be in a human-readable format per the third paragraph, but it's not a requirement. WCAG has a section on reporting conformance, and 2.1 adds a recommendation for metadata, for example.

laudrain:
I may be influenced by Ace where results are in a report.json file and in a human-readable format report.html.
Your prose says "the report be included as a resource" and I am wondering if it would be the json or the html file. May be none of the 2 but still another specific and ad'hoc HTML file ?

mattgarrish:
You could put the Ace HTML report in, but I'm not sure how useful it would be to the average person (json even less so). It's more a validation report with some extra info for evaluators.

I'm thinking more the kind of conformance reports that come out of another tool we've built, which indicate conformance to the EPUB specification, break down the WCAG/EPUB criteria, list the accessibility metadata, etc.

We don't have to enforce the format of the reports, though, and shouldn't. It's better to leave flexibility as the key is providing the information. A WCAG evaluation report would suffice, and there's isn't a strict format on how you create those.

If we need machine processing of the accessibility metadata, we also need to determine who benefits from this most and where it is most likely to get harvested. For unpackaged publications, I still see search engines as the most useful consumer, so having it in the manifest isn't the optimal approach. It could be that we require the discovery metadata in a common record (ideally never onix on the open web) and still include a link in the manifest for finding a report page.

But we don't have to solve all these issues with this PR. This was a response to @iherman asking if we could come up with some better purpose/wording for the current accessibility section. If this doesn't work for now, what suggestions do people have for what the accessibility property is?

mattgarrish:
I've added the following editor's note:

> Machine-readable accessibility metadata may be recommended in whatever format is used to externalize publication metadata (e.g., to ensure availability for search). Depending how this externalizing is done, adding machine-processable accessibility metadata to such a record could take precedence over, or complement, the accessibility record.

mattgarrish:
> My main concern with this approach is the chicken and egg situation caused by having the evaluation report about the publication part of the same publication.

Is there one in the unpackaged form, though? The recommendation is to include a link when a report exists, so a publication can begin life without one and be updated later (it was an intentionally weak recommendation for this reason).

If we try to enforce the presence, then I agree we could have problems.

Separating the discovery metadata is fine with me. The only reason I mentioned it is so we keep it on the radar and because it is possible for an evaluation report to also identify these properties. I don't know that it's optimal myself.

iherman:
@mattgarrish any reasons why this PR is still open?

mattgarrish:
I don't recall we ever discussed it on a call, so it fell off the radar. I think it's fine to merge and continue discussions on it, though.

PR #172, opened 2018-04-05
Update index.html

laudrain:
identifiers and/or addresses allow to retrieve the WP, why representation?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/172.html" title="Last updated on Apr 5, 2018, 6:21 AM GMT (b8b7b24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/172/1af873d...b8b7b24.html" title="Last updated on Apr 5, 2018, 6:21 AM GMT (b8b7b24)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -464,7 +464,7 @@
 						consumption of the content, (e.g., provide a table of contents or a link to one).</p>
 
 					<p>The availability of the address does not preclude the creation and use of other identifiers
-						and/or addresses to retrieve a representation of a <a>Web Publication</a>, whether in whole or
+						and/or addresses to retrieve a <a>Web Publication</a>, whether in whole or
 						in part.</p>
 
 					<div class="note">The Web Publication's address can also be used as value for an identifier link


mattgarrish:
FYI, I removed the emoji that's breaking the diff from the first comment in issue #36, but it looks like the program is building from a cached snapshot.

Removing "representation of" looks fine to me, but I believe @iherman authored this section so I'm hesitant to approve. Was this written when we were thinking there might be different ways of crafting web publications? (html or json manifests, for example)

iherman:
@mattgarrish @laudrain I am fine with the removal. Using the term "representation" tries to be overly precise (per Web Architecture) but is not necessary here imho. (I do not know whether it is indeed I who wrote it...)

dauwhe:
I'm fine with the change, but what function does this paragraph serve? It's saying that a WP might have more than one URL.

1. Does this paragraph have any normative force? This doesn't seem testable. 

2. Is it a good idea for a WP to be available at multiple URLs? If so, is this something that needs to be discussed in the spec, or are we telling people how to design servers, business models, etc.?

mattgarrish:
> Does this paragraph have any normative force? This doesn't seem testable.

I don't think the paragraph has much practical value, to be honest. We don't normatively restrict a WP to one URL, so we're sort of affirming the reality of the web. It looks like it should also be a note, especially the way its positioned between two other notes.

> Is it a good idea for a WP to be available at multiple URLs?

I don't see how this could be prevented. As I read the statement, all I see it saying is:
1. it's possible to retrieve the web publication without going through the address (e.g., from one of the constituent resources)
2. there may be many paths to the address via different urls (e.g., the output of a url shortener; redirects on alternate addresses that are more human-friendly to remember; DOIs and the like).

Whether it's necessary to say this is debatable.

dauwhe:
We seem to have consensus on the smaller issue of the language. This was discussed a bit during the PBG steering committee meeting, so I'm comfortable merging this. We can open a new issue on the normative value of this paragraph.

PR #195, opened 2018-05-14
Rename entry point to entry page

BigBlueHat:
Simple search and replace alteration.

Entry page is what most people have been saying in GitHub issues for weeks now in place of entry point. Changing to match both clarify and match real usage.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/195.html" title="Last updated on May 14, 2018, 1:34 PM GMT (6cffbfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/195/543e227...6cffbfb.html" title="Last updated on May 14, 2018, 1:34 PM GMT (6cffbfb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -405,7 +405,7 @@
 
 					<p>A <a>Web Publication's</a>
 						<dfn>address</dfn> is a <abbr title="Uniform Resource Locator">URL</abbr>&#160;[[!url]] that
-						represents the primary entry point for the Web Publication. If this URL does not resolve to an
+						represents the primary entry page for the Web Publication. If this URL does not resolve to an
 							<abbr title="Hypertext Markup Language">HTML</abbr> document&#160;[[!html]], user agents
 						SHOULD NOT provide access to it to users.</p>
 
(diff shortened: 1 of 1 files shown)


mattgarrish:
Does this mean it's illegal for the address to be anything but HTML?

The entry point/page is not necessarily the same as the one required HTML manifest-linked resource, so the other question is whether we're changing that, too?

I personally find what we have right now confusing, but we never did reach consensus on what the entry has to be, or whether it has to be part of the publication. I believe that's how we ended up with the non-committal "entry point" name.

(I'm fine with the name change, fwiw, but I'd like to clean these other bits up, too, if this is the direction we're moving in.)

BigBlueHat:
@mattgarrish I agree there's much more to clean up, clarify, and/or correct. My hope with this PR was to simply make the spec text match what people are saying in the issues currently.

iherman:
I propose to merge this asap, to avoid merge issues later. @mattgarrish ?

mattgarrish:
I'm fine merging this now so long as we get to the other inconsistencies later... :)

Issue #198, opened 2018-05-15
Is an exhaustive "resource list" required to create a Web Publication?


BigBlueHat:
This was another topic which surfaced during the #193 discussions.

@HadrienGardeur [posed some concerns](https://github.com/w3c/wpub/issues/193#issuecomment-389163035) to a ["dependency gathering" approach](https://github.com/w3c/wpub/issues/193#issuecomment-389163035) proposed by @BigBlueHat.

There are several potential consumption scenarios which should be considered:
#### Current:
   - reading through a Web browser (i.e. follow a link to a publication address)
   - caching and/or offlining the publication in a browser
#### Future:
   - creation of a packaged publication (see https://github.com/w3c/pwpub/)

---

@HadrienGardeur's concerns to the gathering process are below...
> -   conceptually, this means that the bounds of the resources are limited to the reading order (there's no way to indicate that a resource is part of the publication but not in the reading order anymore)
> -   this could trigger the download of very large resources (HD videos) or resources that are useless for rendering (analytics scripts) that would otherwise have been excluded by the author when caching and/or packaging a publication
> -   to cache or package a publication, you need to render every single resource from the reading order, which is going to be very slow and CPU+memory intensive (we've experimented with background rendering in Readium-2 and limit things to only 3 resources)
> -   not all UAs may be able to intercept network requests to cache or package them, this would exclude native mobile apps for example from supporting WPs properly
> -   UAs won't be able to do intelligent preloading on their own (for instance by loading fonts in cache in advance)

BigBlueHat:
> conceptually, this means that the bounds of the resources are limited to the reading order (there's no way to indicate that a resource is part of the publication but not in the reading order anymore)

The reading order (however it's created/expressed) points to the primary resources, which in turn state what resources they depend upon. so, the way you would "indicate that a resource is part of the publication" is by referencing it from a primary resource (i.e. `<img src="">`, `<video>...`, `<link rel="stylesheet">`).

The default consumption of the publication within a browser "just works" (as this is how the browser do things now--there's no "resource list" for a web site you visit, just pages referencing their dependencies).

[...]

PR #200, opened 2018-05-21
Reordered & relisted requirements to match prose

BigBlueHat:
- resorted and reworded Recommended Descriptive Properties
- moved Reading Progression up to Recommended Descriptive Properties
- moved Resource List into Recommended Structural Properties

All changes made to these list match currently merged prose text
about the same requirements.

Related (in part) to conversations in #198. Depending on which area of the spec you were referring to the "resource list" was either required or recommended. This PR makes the list match the prose--i.e. fixes that bug in the requirements list.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/200.html" title="Last updated on May 21, 2018, 8:38 PM GMT (ccd0138)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/200/b72ab19...ccd0138.html" title="Last updated on May 21, 2018, 8:38 PM GMT (ccd0138)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -372,8 +372,8 @@
 					</dd>
 					<dt>Structural Properties</dt>
 					<dd>
-						<p>REQUIRED: <a href="#wp-default-reading-order">default reading order</a></p>
-						<p>RECOMMENDED: <a href="#wp-resource-list">resource list</a> and
+						<p>RECOMMENDED: <a href="#wp-default-reading-order">default reading order</a>,
+								<a href="#wp-resource-list">resource list</a>, and
 								<a href="#wp-table-of-contents">table of contents</a></p>
 					</dd>
 				</dl>


BigBlueHat:
[Default Reading Order](https://w3c.github.io/wpub/#wp-default-reading-order) is actually not currently required according to the prose ("If a user agent requires a default reading order..."). Consequently, I'm adding a separate commit to move "Default Reading Order" down to Recommended to match the prose.

mattgarrish:
> is actually not currently required according to the prose

That prose was inserted for error handling. The "if" is not making the reading order non-required, but acknowledging that a user agent might do something else if there isn't a default reading order (disable sequential navigation, bail out on the publication, ...).

Goes back to the discussions we had about whether its enough to say properties are required without also saying what should happen when they are omitted.

I agree it's confusing, but I'm not a big fan of making user agents construct things the author forgot (at least not with the latitude given). If we want consistency, the answer here is probably that the user agent disable any progression features when a default reading order isn't specified.

HadrienGardeur:
@mattgarrish I would suggest a slightly different solution: if the author forgot to include a default reading order, the UA defaults to a default reading order where the WP address is the only item listed.

Right now, the WP address and the default reading order are the only two requirements that we have.

mattgarrish:
@HadrienGardeur Sure, any approach that is going to be consistent across user agents is fine with me. I understand the desire to have user agents fill in the blanks, but we should try much harder to ensure that the experience is consistent.

Titles are another case, where depending on the user agent you use you might discover the same publication has different names if one isn't specified in the infoset.

iherman:
@mattgarrish merge this, too (before the F2F)?

mattgarrish:
I believe @BigBlueHat still has to make an update before we merge.

iherman:
@BigBlueHat there is now a conflict with the main branch (I guess due to the merged PR of @HadrienGardeur), and I would expect some conflicts with the other PR I have just filed (#211). I wonder whether it is not quicker if you looked at #211 and possibly made some changes/commits on that one, rather than chasing the conflicts in this PR (which has become outdated...)

iherman:
B.t.w., the F2F resolution is analogous to https://github.com/w3c/wpub/pull/200#issuecomment-390912983 of @HadrienGardeur, and is part of #211 .

PR #213, opened 2018-06-05
Manifest rewrite

iherman:
I tried to provide a skeleton for the manifest section based on the resolutions in Toronto. Apart from the content that is only partially provided, the question is also editorial, ie, whether this editorial approach would work (or not).

Cc: @GarthConboy @TzviyaSiegman @mattgarrish

### DIFF
--- a/index.html
+++ b/index.html
@@ -794,7 +794,7 @@
 
 					<p>
 						<a href="infoset-meta">Desciptive Properties</a> in the Web Publication Manifest are based on the terms defined by <a href="https://schema.org">Schema.org</a>&nbsp;[[schema.org]] (including <a href="http://schema.org/docs/schemas.html">hosted extensions</a> of Schema.org).
-						This means that the descriptive infoset properties are mapped onto one or several Schema.org properties (inheriting their syntax and semantics).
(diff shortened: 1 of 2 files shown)


HadrienGardeur:
The WebIDL currently uses `resources` and `reading_order`. Could we align with these terms for now instead of `publ-resources`?

index.html
```diff
@@ -770,16 +770,117 @@ <h3>Extensibility</h3>
...
+	}
+	&lt;/script&gt;
+</pre>
+
+				<section>
+					<h4>Descriptive Infoset Properties in the Web Publication Manifest</h4>
+
+					<p>
+						<a href="infoset-meta">Desciptive Properties</a> in the Web Publication Manifest are based on the terms defined by <a href="https://schema.org">Schema.org</a>&nbsp;[[schema.org]] (including <a href="http://schema.org/docs/schemas.html">hosted extensions</a> of Schema.org).
+						This means that the descriptive infoset properties are mapped onto one or several Schema.org properties (inheriting their syntax and semantics).
```
TzviyaSiegman: s/mapped onto/mapped to

Issue #139, opened 2018-02-13
WP affords guided navigation


TzviyaSiegman:
[2.2.1 Default Reading Order](https://www.w3.org/TR/pwp-ucr/#reading-order)
There should be a means to indicate the author’s preferred navigation structure among the resources of a Web Publication.

picking up on #52 
This is addressed in the current draft in the info-set section about [structural properties](https://w3c.github.io/wpub/#infoset-structure), which currently has:
* Default reading order (required)
* Resouce list (required)
* Table of contents (recommended)

HadrienGardeur:
In the draft this is also handled in two different sections, specifically about affordances:

- [Reading Order](https://w3c.github.io/wpub/#aff-reading-order)
- [Table of Contents](https://w3c.github.io/wpub/#aff-toc)

HadrienGardeur:
It's also worth pointing out that in the context of EPUB, we mostly used the term "guided navigation" to panel by panel navigation for comics.

It's unclear at this point if this group will also address this requirement (or if we'll have an equivalent at all for FXL).

jmulliken:
Could this be combined with TOC or Reading Order by saying WP affords automated reading order if no prescribed navigation is provided?

WSchindler:
Affordance name: **Guided navigation (#139)**
 
**1. Short description:** 
The author should be able to define a preferred navigation across the (primary?) resources of a Web 
Publication. 
From the perspective of a UA this would be the **default reading order** which defines the sequence in which to present the resources of the Web Publication to the user. A **Table of Contents** may expose that order explicitly to the user.

**2. Dependencies:** 
The preferred navigation would be defined in terms of a _**default reading order**_ which must be a required item of the manifest. It is based on a list of WP resources which contains all internal resources and may also list external resources. This **list of resources** should as well be a required item of the manifest.

_aside: The UA needs to be able to quickly identify and locate the resources that are part of the logical unit of a WP. The preferred navigation would typically only define an ordering of the major building blocks/resources -  however we dub these. I would suppose that references to external resources that are not part of WP would not be part of the preferred navigation._ 

**3. Use case Reference:**
_Web Publications Use Cases and Requirements: W3C Interest Group Note 02 May 2017_
**2.2.1 Default Reading Order**
[https://www.w3.org/TR/pwp-ucr/#reading-order](url)

Req. 12: There should be a means to indicate the author’s preferred navigation structure among the resources of a Web Publication.
[...]

PR #214, opened 2018-06-08
Added text for the affordance "Guided Navigation"

iherman:
This was originally provided in #139, by @WSchindler; text in https://github.com/w3c/wpub/issues/139#issuecomment-385930513 essentially taken over with some editorial changes.

This PR is the first that uses a submitted affordance in the document; it should also be looked whether it is fine in the editorial structure of the document. Something definitely for @mattgarrish ...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/214.html" title="Last updated on Jun 8, 2018, 11:27 AM GMT (dfb8e81)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/214/a895843...dfb8e81.html" title="Last updated on Jun 8, 2018, 11:27 AM GMT (dfb8e81)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1279,8 +1279,57 @@
 
 					<p class="issue" data-number="145"></p>
 				</section>
-			</section>
 
+				<section>
+					<h4>Guided Navigation</h4>
+					<p class="issue" data-number="139"></p>
+
+					<section>
+						<h5>Short description</h5>
+						<p>
+							The author should be able to define a preferred navigation across the resources of a Web Publication. From the perspective of a User Agent this would be based on the <a>Default Reading Order</a> which defines the sequence in which to present the resources of the Web Publication to the user. A <a>Table of Contents</a> may expose that order explicitly to the user.
+						</p>
+					</section>
+
+					<section>
+						<h5>Dependencies</h5>
+						<p>
+							The preferred navigation would be defined in terms of the <a>Default Reading Order</a> which is a required item of the Information Set.
+						</p>
+					</section>
+					<section>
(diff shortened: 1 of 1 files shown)


iherman:
Specific questions: the affordance includes extra use case examples. Wouldn't it be better to

- add those examples to the use case document (cc @atyposh , @llemeurfr )
- the affordance text in the WP draft should only use _references_ to the use case document, rather than reproducing them in this document

(Just worried about the bloating of the document overall...)

HadrienGardeur:
I find our current usage of "guided navigation" confusing:
- it seems to be basically moving forward/backward in the reading order based on this PR
- not clear how different this is from [section 7.3.1](https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/214.html#aff-reading-order)
- in EPUB, guided navigation was primarily used in the context of comics/magazines, where resources from the default reading order where referenced but with [the addition of specific media fragments](http://www.idpf.org/epub/renditions/region-nav/#sec-3.5.1)
- IMO, this is where we should draw some inspirations, as an alternate way of navigating the publication (not necessarily always based on media fragments)

mattgarrish:
I agree with Hadrien that this appears to be a duplication of the current Progression section.

I probably need more context before judging the merits for inclusion, but it reads more like a definition of how to write a section like the current Progression one (e.g., what to consider and where it comes from) than something that I'd probably expect to find verbatim in a specification.

iherman:
Decided to close it on the call 2018-06-25 without merge (the document will be restructured)

Issue #212, opened 2018-06-05
"order" in the reading order in JSON-LD


iherman:
Care should be taken that, in JSON-LD, a JSON 'array' does not necessarily means strict order (this is an ugly side effect of… don't ask:-). This means that we will have to be careful how to define the reading order in terms of JSON-LD.

iherman:
This is just a placeholder issue, in fact: it is necessary to use the `@list` feature of JSON-LD when defining the relevant term (is it "spine"?) in the Web Publication Manifest context file.

llemeurfr:
This is another incentive to create our own context, and use `"@container": "@list"`in the definition of the reading order property. This will avoid overloading the syntax of the property in each instance via `@list`. 

Ref. https://json-ld.org/spec/latest/json-ld/#sets-and-lists

dauwhe:
So we can't do this...

```json
{
    "spine":  ["a.html", "b.html", "c.html"]
}
```

...but we have to do this...

```json
{
    "spine": {
      "@list": ["a.html", "b.html", "c.html"]
    }
}
```
unless we add a context...

```json
{
  "@context": {
    "spine": {
      "@container": "@list"
      }
    },
    "spine":  ["a.html", "b.html", "c.html"]
}
```

...even though most implementations would process structural metadata as JSON rather than JSON-LD? This sounds like complexity that doesn't serve the user. I just hope I don't have to write the blog post explaining why arrays are ordered in JSON but not JSON-LD :)

akuckartz:
@dauwhe Well, the strange thing is that sets are always ordered in JSON. JSON-LD correctly handles (unordered) sets :-)

HadrienGardeur:
Is there a good place to start drafting this WP context document? Directly in this repo?

Right now, I think that we need to:

- define both `resources` and `reading_order` (I'm surprised to see `spine` again, we used it in EPUB and RWPM but I thought that the term would disappear)
- indicate that `reading_order` is a list using `"@container": "@list"`
- maybe map `reading_order` to a schema.org term? (http://schema.org/hasPart)

iherman:
I would think we could put it into the "common" folder of this repository. I will then have to put up a redirection in the /ns directory from the 'canonical' version of the context file's URL to this one. But that is not urgent, probably.

iherman:
I believe by virtue of #215 this issue can be closed now.

PR #215, opened 2018-06-08
First attempt at crafting a JSON-LD context

HadrienGardeur:
Based on our discussion in #212 here’s a first draft of the context document.

### DIFF
--- a/common/context.jsonld
+++ b/common/context.jsonld
@@ -1,12 +1,5 @@
 {
   "@context": {
-    "bib": "http://bib.schema.org/",
-    "artist": "http://bib.schema.org/artist",
-    "colorist": "http://bib.schema.org/colorist",
-    "inker": "http://bib.schema.org/inker",
-    "penciler": "http://bib.schema.org/penciler",
-    "letterer": "http://bib.schema.org/letterer",
-    "readBy": "http://bib.schema.org/readBy",
     "reading_order": {
       "@container": "@list",
       "@id": "http://schema.org/hasPart"
--- a/common/context.jsonld
+++ b/common/context.jsonld
@@ -1,0 +1,18 @@
+{
+  "@context": {
+	"bib": "http://bib.schema.org/",
+	"artist": "http://bib.schema.org/artist",
+    "colorist": "http://bib.schema.org/colorist",
+    "inker": "http://bib.schema.org/inker",
+    "penciler": "http://bib.schema.org/penciler",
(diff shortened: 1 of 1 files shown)


iherman:
My reading of, eg, the example at https://bib.44.20180402t222137.schemaorgae.appspot.com/colorist but also the fact that the original schema.org json context has a "@vocab" set to "schema.org" is that even terms originating from the "bib" extension of schema.org are represented by URLs in the "schema.org" space. In other words, the mapping of, say, "colorist" is to be "http://schema.org/colorist". 

If that is correct then:

- our context file should not include anything on "colorist"
- we should require (as it is in the draft) to include the schema.org context file into the manifest, which would provide the right "@vocab"

Am I missing something?

iherman:
We will have some enjoyable bikeshedding on setting the terms:-)

llemeurfr:
Can we really remove refs to bib.schema.org?
... http://schema.org/colorist leads to "The term 'colorist' is not in the schema.org core vocabulary, but is defined in an extension: bib: colorist"
... and http://bib.schema.org/colorist leads to the proper page.

llemeurfr:
ok, I see that now, and in the bib page the note: "extension terms can be used in schema.org markup in the normal manner; it is not necessary for markup publishers to indicate which extension a term is currently in. Terms may move between extensions over time (e.g. from pending to the core) without the need for the corresponding markup to change."

iherman:
> 
> Can we really remove refs to bib.schema.org <http://bib.schema.org/>?
> ... http://schema.org/colorist <http://schema.org/colorist> leads to "The term 'colorist' is not in the schema.org <http://schema.org/> core vocabulary, but is defined in an extension: bib: colorist"
> ... and http://bib.schema.org/colorist <http://bib.schema.org/colorist> leads to the proper page.
> 
> 

That is correct but that is not what counts: if you look at the JSON-LD example on that page, they only use the schema.org <http://schema.org/> context file, ie, the term is mapped onto "http://schema.org/colorist <http://schema.org/colorist>". it is a bit weird, I agree, but not our decision…

(B.t.w., there is now a redirection of https://www.w3.or/ns/wpub.jsonld <https://www.w3.or/ns/wpub.jsonld> to the github version. It is temporary, ie, at some point it should be a hard content.)

HadrienGardeur:
@iherman thanks for the redirection. 

Tried a few more examples with that and it looks like we need to add an `@id` for `resources` as well, otherwise it gets rejected by the JSON-LD playground.

Issue #220, opened 2018-06-11
Setting the (text) direction


iherman:
This is a completely open issue at this moment, both for JSON-LD and Schema.org... The only (incomplete) approach would be to rely on, and base everything, on the UTF-encoding of the text...

iherman:
See also the separate discussion on the JSON-LD 1.1 CG: https://github.com/json-ld/json-ld.org/issues/583

iherman:
Another reference: https://w3c.github.io/string-meta/

danielweck:
> The only (incomplete) approach would be to rely on, and base everything, on the UTF-encoding of the text...

Do you mean Unicode Bidi? http://unicode.org/reports/tr9/

iherman:
@danielweck 

More specifically, see https://github.com/json-ld/json-ld.org/issues/583#issuecomment-364141212 : which referred to this. The following discussion (which was, as far as I am concerned, inconclusive) gave some pro and cons to that approach.

Note that the JSON-LD CG decided to defer that issue to the JSON-LD WG which has just been formed; I hope that the discussion will re-start with some more people involved (eg, Schema.org people as well). We may want to defer this issue to see where that discussion go.

iherman:
@TzviyaSiegman reminded me that there is another approach that is perfectly viable, namely use the HTML datatype. What this means in practice is that, if a text has bidirectional issues, it could use HTML syntax and the result would be considered to be a string of an HTML datatype in RDF parlance. Here is what it would mean in JSON-LD:

```
{
    "name" : {
        "@value" : "We find the phrase '<span dir="rtl" lang="he">פעילות הבינאום</span>' 5 times on the page.",
        "@type" : "rdf:HTML"
    }
}
```

(The trick is to ensure that the character '5' appears on the right hand side of a Hebrew text. If the span is not used, the number will be used as if it was part of the hebrew text and will appear on the _left_ of it!)

From an internationalization point of view, that is much better, because it gives a better control. We _could_ therefore say that, for example for the term `name`, the author should use that approach. I see two problems with that, too:

1. It may be an extra load on authors (and maybe not; I am not sure how frequent these occurrences are)
2. Just as for the case #219, while the above JSON-LD is perfectly fine, the google structured data tester does not accept it :-(

laudrain:
When written as:
```
{
    "name" : {
        "@value" : "We find the phrase '<span dir='rtl' lang='he'>פעילות הבינאום</span>' 5 times on the page.",
[...]

Issue #219, opened 2018-06-11
Expressing the language of an item, say, "name"


iherman:
The requirement is that it should be possible to express the language of, say, the title or name of the author individually. This does not seem to work in Schema.org...

iherman:
In pure JSON-LD, the following idiom is acceptable and produces the right result:

```
{
  "@context":"http://schema.org",
  
  "name" : {
    "@value": "Karinthy Frigyes",
    "@language" : "hu"
  }
}
```

However, the same example is rejected by the [Google tester](https://search.google.com/structured-data/testing-tool). Indeed, `name` has the value of `Text`, ie, only a text value, and the structure above is not accepted.

Maybe discuss it with the schema.org maintainers...

HadrienGardeur:
But from a JSON-LD perspective, what you've expressed should be just `Text` with the language set to `hu`. Is this truly a problem with schema.org rather than Google's ability to support JSON-LD correctly?

iherman:
@HadrienGardeur: I do not know. As I said, to be discussed with them.

At the moment, I included that type of structure in #221 , but flagged it with this issue.

HadrienGardeur:
It's also worth pointing out that with a context document, we could use `"@container": "@language"` on various terms.

Instead of:

```
"name" : {
  "@value": "Karinthy Frigyes",
  "@language" : "hu"
}
```

... you could write: 

```
"name" : {
  "hu": "Karinthy Frigyes"
}
```

iherman:
In JSON-LD: yes. But is that acceptable in Schema.org: I do not believe so...

HadrienGardeur:
Once again, the issue is with Google's ability to work with JSON-LD, not with schema.org terms.

iherman:
This is now part of the draft, and schema.org is looking at this. Closing this issue.

Issue #218, opened 2018-06-11
Expressing the overall language for the Metadata


iherman:
The right approach is to _extend_ the context through

```
{
   "@language" : "fr"
} 
```

in the (list of) `@context`. Although not rejected by schema.org, it is also ignored.

iherman:
Probably should be discussed with Schema.org

iherman:
Has been covered via the latest draft (2018-07-25). Closing

Issue #217, opened 2018-06-11
Order sensitivity of terms like 'creator', 'author', ...


iherman:
There is no order sensitivity for many items in Schema.org. It is possible to use a JSON-LD array for, say, the author, but the order is not preserved (technically, the mapping is on a set of statement, not a list)

iherman:
Has been added via #232; closing

PR #221, opened 2018-06-11
First attempt for Schema.org mapping

iherman:
Following the discussions at the F2F, I tried to map the descriptive items to schema.org. By doing so, I have identified some issues (#217, #218, #219, #220) although some of those were already known just not yet recorded.

I have made some slight changes compared to the [wiki page](https://github.com/w3c/wpub/wiki/Descriptive-Infoset-Properties-vs.-Schema.org-table):

- The mapping on `inLanguage` would be semantically wrong. That term refers to the language of the publication (eg, Book); we explicitly say that all language settings are meant _for the metadata itself_. Ie, we have to fall back on the JSON-LD features setting the languages, though this raise its own problem (see #219)
- the table refers to '@id', but there is also a 'id' in schema.org that maps against '@id'. I would think that is a better match.
- I created a table for a bunch of 'creators'; I am not sure it is complete and it is a good idea to list them all here...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/221.html" title="Last updated on Jun 13, 2018, 8:42 AM GMT (4fe82e0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/221/86f437e...4fe82e0.html" title="Last updated on Jun 13, 2018, 8:42 AM GMT (4fe82e0)">Diff</a>

### DIFF
--- a/common/context.jsonld
+++ b/common/context.jsonld
@@ -5,7 +5,11 @@
       "@id": "http://schema.org/hasPart"
     },
     "resources": {
-      "@container": "@set"
+      "@container": "@set",
+      "@id": "http://www.w3.org/ns/wpub/resources"
+    },
+    "readingProgression": {
+        "@id": "http://www.w3.org/ns/wpub/resources"
     },
     "author": {
         "@container" : "@list",
--- a/common/context.jsonld
+++ b/common/context.jsonld
@@ -1,16 +1,5 @@
 {
     "@context": {
-        "readingOrder": {
-            "@container": "@list",
-            "@id": "http://www.w3.org/ns/wpub#readingOrder"
-        },
-        "resources": {
-            "@container": "@set",
-            "@id": "http://www.w3.org/ns/wpub#resources"
-        },
-        "readingProgression": {
-            "@id": "http://www.w3.org/ns/wpub#readingProgression"
-        },
         "author": {
             "@container": "@list",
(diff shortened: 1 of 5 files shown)


iherman:
> @HadrienGardeur requested changes on this pull request.
> 
> Looks good but there seems to be a mismatch between the list of contributors listed in the draft vs context (much shorter list).
> 

Yes, indeed. I wanted to see whether everybody is all right adding all these terms to the document in the first place before filling up the jsonld context file. If everybody is fine with the approach as a whole, I will complete it.

> There's also a typo in the context document for readingProgression:
> 
> "readingProgression": {
>   "@id": "http://www.w3.org/ns/wpub/resources <http://www.w3.org/ns/wpub/resources>"
> },

Oops. Copy/paste error; handled it.

iherman:
@HadrienGardeur I have update the context file with the complete list of 'creators'.

iherman:
@TzviyaSiegman @HadrienGardeur @mattgarrish @GarthConboy @llemeurfr : I have added the structural elements to the manifest, which means that the first draft of the manifest, for those terms that we have in the infoset, is almost there.

What is missing is what exactly we do with cover pages and/or privacy policies. This is a separate issue...

TzviyaSiegman:
Once conflicts are resolved, please merge this PR

Issue #223, opened 2018-06-11
Other navigation elements beyond ToC


HadrienGardeur:
This is based on my gap analysis between EPUB 3.2 and WP: https://github.com/w3c/wpub/issues/176#issuecomment-382634697

EPUB provides a number of additional lists that can all be expressed in the Navigation Document such as:
- [page list](https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-nav-pagelist)
- [landmarks](https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-nav-landmarks)
- [list of audio files, list of tables, list of videos and list of images](https://idpf.github.io/epub-vocabs/structure/structure.html#navigation)

Do we need any of these additional elements in WP?

HadrienGardeur:
Page list seems to be very high on the list of requirements for education, it's probably worth considering in both WP and EPUB4.

laudrain:
It's also a must have for a11y for all kind of books in general.

danielweck:
Re. list of page breaks (aka page list): well, @laudrain "for all kind of books in general" **that have an equivalent print representation** (just to be clear that this is specifically about allowing content creators to encode references to "printed pages", nothing to do with "page numbers" generated in the context of dynamically-paginated reflowable documents, i.e. the reading system's own pagination capabilities).

GarthConboy:
Seems, at least for page-list, this could be a manifest item pointing to an ARIA-tagged page-list <nav> element, much like the TOC.

avneeshsingh:
Indeed, adding pagelist should be straight forward because we already have doc-pagelist in DPUB aria roles.
From accessibility point of view, I will also support pointing to other important pieces in the publications like landmarks, notes etc. Quick navigation to important chunk of information is very helpful for people with disabilities.

GarthConboy:
+1 to Anveesh above.   The "page list" should be an ARIA tagged "doc-pagelist" tagged `<nav>` element.  Such should be pointed to from the Manifest (and tagged as such).  The `<nav>` element could well be in the Entry page, though not required.

Such an approach would work a desired in a non-WP aware UA/RS (assuming the browser encounter the so-labeled `<nav>` (in the Entry page or otherwise).

 I'd tend to think a link to the page-list would also be in the TOC, but that's clearly not required.

BigBlueHat:
If the page list is not included in the entry page, then how is the human reader to find it (since they can't consult the manifest)?

[...]

PR #224, opened 2018-06-11
Aligned WebIDL proposal with recent discussions

HadrienGardeur:
Based on the F2F and post-F2F calls, I’ve aligned the WebIDL a bit more with schema.org.

This PR also contains a first take on handling the TOC in the WebIDL, using a tree structure and a single WebIDL dictionary.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/224.html" title="Last updated on Jun 12, 2018, 2:35 PM GMT (5b9d711)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/224/86f437e...5b9d711.html" title="Last updated on Jun 12, 2018, 2:35 PM GMT (5b9d711)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1518,7 +1518,7 @@
 				<pre class="idl">
                     dictionary PublicationLink {
                         required    DOMString                           url;
-                                    DOMString                           type;
+                                    DOMString                           fileFormat;
                                     DOMString                           name;
                                     sequence&lt;DOMString&gt;           rel;
                                     sequence&lt;PublicationLink&gt;     children;
@@ -1533,7 +1533,7 @@
 					</dt>
 					<dd>Contains the URL of the linked resource.&#160;[[!url]]</dd>
 					<dt>
-						<dfn>type</dfn>
+						<dfn>fileFormat</dfn>
 					</dt>
 					<dd>Contains the MIME media type of the linked resource.&#160;[[!rfc2046]]</dd>
 					<dt>
--- a/index.html
+++ b/index.html
@@ -1370,17 +1370,17 @@
 
 				<pre class="idl">
                     dictionary WebPublicationManifest {
-                        required    DOMString                               address;
+                        required    DOMString                               url;
                                     DOMString                               lang;
-                                    TextDirection                           dir = "auto";
-                                    TextDirection                           progression = "auto";
-                                    sequence&lt;LocalizableString&gt;       title;
-                                    DOMString                               identifier;
+                                    TextDirection                           direction = "auto";
+                                    TextDirection                           readingProgression = "auto";
+                                    sequence&lt;LocalizableString&gt;       name;
+                                    DOMString                               id;
                                     sequence&lt;Contributor&gt;             authors;
-                                    DOMString                               modified;
-                                    DOMString                               publication_date;
+                                    DOMString                               dateModified;
+                                    DOMString                               datePublished;
                                     sequence&lt;PublicationLink&gt;         links;
-                                    sequence&lt;PublicationLink&gt;         reading_order;
+                                    sequence&lt;PublicationLink&gt;         readingOrder;
                                     sequence&lt;PublicationLink&gt;         resources;
                                     sequence&lt;PublicationLink&gt;         toc;
                     };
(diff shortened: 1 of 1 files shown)


HadrienGardeur:
For now, I'm still using `type` instead of `fileFormat` (http://schema.org/fileFormat) in `PublicationLink`, which is the core element used in the default reading order, the list of resources and the table of contents.

Should we align on this term as well?

iherman:
@HadrienGardeur :

- Yes, I believe it would be better to align the 'type' terms as well, just for the purpose of consistency
- I was wondering yesterday whether 'name' or 'description' is the proper term to use here (both are possible). The examples that come to my mind are all for a link that require, eg, an alt text for a11y, in which case 'description' may be more appropriate...
- I do not remember having talked about 'children', and I am not sure we will need this. The values used in resources or readingOrder are all a single layer list, no hierarchies there.  
The only reason I could see is for the purposes of hierarchical TOC-s. But are we sure we would use the same structures for the representation of TOC?

iherman:
(Admin remark: I wonder whether it wouldn't be better to merge this PR with #221) as on bigger PR; the two are fairly closely related... Or merge them into the master pretty much one after the other.

@TzviyaSiegman @GarthConboy , what do you think?

HadrienGardeur:
@iherman when I first drafted the WebIDL, I mostly thought about this structure as a link.

For me, a link is a more natural fit than similar structures in schema.org.

In HTTP, HTML, Atom or even EPUB, the link element remains mostly the same thing with only small differences.
That's why the initial WebIDL had `href`, `rel`, `title` and `type`.

If we align with the schema.org terms, then I agree that we should fully align. That said, these terms feel more awkward to me than what we had before: I prefer `title` over `name` and `type` over `fileFormat`.

Right now, this structure is used for:

- publication-level links
- the default reading order
- the list of resources
- the table of contents

We could have separate structures for each of these elements, but I don't think there's a strong case for it.

The only differences between `readingOrder`/`resources` and `toc` are:

- we only need `children` for `toc` (plus similar navigation lists being discussed in #223) 
- we might not need a `rel` for `toc` (but it could be useful)
- `name` might or might not be required for the `toc`

For the rest:

- `url` remains required
- `type` remains optional

I would prefer keeping a single structure for now and only divide things into two different structures once we have a strong case for it.

iherman:
@HadrienGardeur o.k. let us have one structure on the WebIDL level. (And yes, I was/am also a bit uncomfortable using `name` for what is a title, for example, but we turned out to be happy with it at the F2F discussion, so we should run with this now.)

That being said, if I translate this to JSON-LD structures, adding the children may be unnecessary. On the manifest level the TOC is 'just' a link to an HTML element somewhere, meaning that it does not have any 'structure' on the JSON-LD level.

I will try to find some more time expanding #221 to include that, which would mean defining the reading order and the resources a breeze, except that the issue of where the `rel` values come from may still remain open (would we try to extend that by IANA? would we use URL-s for non-IANA relations).

dauwhe:
> On the manifest level the TOC is 'just' a link to an HTML element somewhere, meaning that it does not have any 'structure' on the JSON-LD level.

I was wondering about this. In the infoset TOC is clearly a link to an HTML table of contents, but in the IDL the TOC member is described as "Contains the table of contents."

HadrienGardeur:
@iherman WebIDL and JSON-LD are not necessarily meant to be the same.

For the ToC specifically, in WebIDL we'll end up with the results of parsing an HTML document.

Regarding rel values:

- I think we should definitely align with the IANA link registry
- URLs for extensions is the usual extension mechanism for rel
- we need to allow for more than a single rel, this has been problematic in Atom for example in the past

iherman:
@HadrienGardeur I have the impression we are violently agreeing:-)

iherman:
@HadrienGardeur looking that schema.org definition of `name` and `description` (which are both very shallow...) I still wonder whether we should not use `description` either additionally or instead of `name`. `description` seems to be a better fit for, e.g., the alternate text description of an image used for a cover (something that came up on the discussion yesterday...).

PR #238, opened 2018-06-21
Restructure to de-dup definitions

mattgarrish:
Despite the branch name, what started as an editorial review turned into a restructuring. I found it complicated to have separate sections for the infoset and manifest, with the manifest continually quoting and referrring back to the infoset, so I started a re-org to combine the two.

This PR takes all the properties and creates a new section for them. Each property begins with its definition and is followed by a subsection that contains the infoset requirements and another with the manifest expression requirements. The section on constructing a web publication is moved up ahead of this section, and will explain the relationship between the infoset and manifest and also contain the preliminary sections on constructing the json.

I'm opening this now as I'd like to merge the new structure -- if everyone is okay with it -- so that the ongoing work doesn't diverge further and require manual patching. There is definitely more editorial work that I need to do even on this revised structure (it ain't perfect), but I plan to do that next.

(We can also look at linking from the properties to the feature(s) that they enable as an additional next clarifying step, as well as back-linking from features to the properties that afford them. The features don't all map 1:1 with a property, so merging might create a mess.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/238.html" title="Last updated on Jun 21, 2018, 4:22 PM GMT (2c8ecdd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/238/8dd6539...2c8ecdd.html" title="Last updated on Jun 21, 2018, 4:22 PM GMT (2c8ecdd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -308,18 +308,23 @@
 								<li><a href="#wp-cover">cover</a></li>
 								<li><a href="#wp-creators">creator</a></li>
 								<li><a href="#wp-language-and-dir">language</a></li>
-								<li><a href="#wp-mod-date">modification date</a></li>
+								<li><a href="#wp-mod-date">last modification date</a></li>
 								<li><a href="#wp-privacy">privacy policy</a></li>
 								<li><a href="#wp-pub-date">publication date</a></li>
+								<li><a href="#wp-reading-progression">reading progression direction</a></li>
 								<li><a href="#wp-title">title</a></li>
 							</ul>
 						</div>
 					</dd>
 					<dt>Structural Properties</dt>
 					<dd>
-						<p>REQUIRED: <a href="#wp-default-reading-order">default reading order</a>, <a
-								href="#wp-resource-list">resource list</a>, and <a href="#wp-reading-progression"
-								>reading progression</a>.</p>
+						<div>
+							<span>REQUIRED: </span>
+							<ul class="flat">
+								<li><a href="#wp-default-reading-order">default reading order</a></li>
+								<li><a href="#wp-resource-list">resource list</a></li>
+							</ul>
+						</div>
 						<p>RECOMMENDED: <a href="#wp-table-of-contents">table of contents</a></p>
 					</dd>
 				</dl>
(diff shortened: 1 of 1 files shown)


mattgarrish:
I don't see the preview option showing up, so just in case it never does here's a rawgit preview: https://rawgit.com/w3c/wpub/d8f0c9bcb7a0523825b67252b32469adfeb3a020/index.html

iherman:
I wonder whether the preview fails because respec reports linking errors...

iherman:
This PR is essential for further work, and we should not submit new PR to the main branch until this is merged, otherwise catching up would become way too complicated for the editors. I would therefore propose to merge this as soon as possible unless some fundamental issues pop up, say before end of business on Friday...

mattgarrish:
> I wonder whether the preview fails because respec reports linking errors...

I wonder if it's because there are too many changes and github doesn't provide a diff file for index.html. At any rate, I've corrected the bad links, but most were broken from before this PR.

mattgarrish:
> I think chapter 4.3 may need some reordering and maybe even some additional structuring.

Are you thinking of a new section of properties? There's no specific ordering right now beyond the earlier agreement to list everything alphabetically for simplicity of lookup.

iherman:
Yes, I believe that the alphanumeric order does not really work in 4.3.

I could imagine something like

4.3 Structural properties
4.3.1. Introduction
4.3.2 Resource lists (not sure about the title)
4.3.2.1 Default Reading Order
4.3.2.2 Resources
4.3.2.3 List of Extra Resources
4.3.3 Resource references (??? I am not sure about the title)
4.3.3.1. Accessibility Report
4.3.3.2. Cover
4.3.3.3. Privacy Policy
4.3.3.4 Table of Content (that is still pending on an issue resolution, but I believe it will end up here)

The idea is that all items in 4.3.3 (plus also some possible items from extensibility) rely on 4.3.2

mattgarrish:
Although... structural properties were originally intended to be properties that defined structures of the publication rather than simply describing some aspect of it.

There's a disconnect for me now in trying to grasp structural as any property that links to a resource. Either the name is now wrong, or maybe we should look at more groupings and perhaps de-emphasize the classifications in the requirements section.

What I see when I look at what we have right now are:

- Structural properties
  - cover
  - default reading order
  - table of contents

- Informative properties
  - accessibility report
  - privacy policy

- Resource enumerations
  - resource list
  - extra resources

iherman:
@mattgarrish I agree that the term 'structural property' is not ideal...

Your categorization does make sense, I agree, from a high level. The problem is how we present this, because, technically, the default reading order is also a resource enumeration, and the others are properties that use one of the resource enumerations as tools the express themselves.

One could say that the default order, the resources and the external resources are all "categorizations" of resources that are used in the WP, and we also define some specific such resources (and the user can add more, like an ONIX file). Would that terminology work?

mattgarrish:
Sure, I wrote long a boring argument about the dual nature (structure and enumeration) and the progression aspect being more important than the enumeration aspect, and then gave up on it. A name like "categorizations" might work as a bridge.

It's never going to be a perfect fit wherever we put it because of the two purposes it serves.

iherman:
@mattgarrish The preview works now! 

Looking at it, I also realize there is another problem: 4.3.1. seems to be a leftover. The `StructuredValue` did not work out after all, hence the need for `PublicationLink` (`StructuredValue` still appears in the intro part of the section and in the example).

iherman:
I have to call it a day now, but I guess by tomorrow morning I will find a new commit with the changes... (note that I may not be that available tomorrow, some works to be done in our home:-(

mattgarrish:
> I thought we had agreed to avoid trying to define new rel values with IANA (e.g. cover-page) if possible?

Not sure. I probably missed that decision... :)

There's still a lot of work to do patching up this re-org, so probably good to start a new issue than solve it here.

> 4.3.1. seems to be a leftover.

I've merged the two former property intros (well, more just crammed the paragraphs together for now). I'll have a look when I do another pass to fix up. There are a number of places in need of massaging.

> I guess by tomorrow morning I will find a new commit with the changes

It's there now. Maybe by tomorrow I'll have some actual editorial review of the text done, not just the structure, but I'll do it in small commits and on what I figure are uncontroversial changes so we can merge this as soon as possible.

iherman:
@TzviyaSiegman I think what we try to avoid is to define new mime types by IANA. I do not see a problem defining new rel values (although we should try to use existing rel values when we can).

iherman:
@mattgarrish I think this should be merged now. As you said, there are some aspects that should be updated, but to be able to possibly share this load, we should have this merged to avoid chaos...

mattgarrish:
Yes, I think I'll go ahead and merge as it doesn't appear to be generating any negative feedback.

Issue #245, opened 2018-06-25
embedded manifest requires id?


mattgarrish:
In order to reference the script from the manifest link, shouldn't an id attribute also be required/recommended?

Right now, the only requirement I can find is that the script element have a specific media type. The section on obtaining a manifest doesn't go into extraction and whether the user agent is expected to parse a document until it finds a script with the specified media type if the resource is HTML and no fragment has been provided in the URL.

iherman:
+1 to have a must. Otherwise the link from the <link> element does not work...

iherman:
Already in the draft, closing.

PR #248, opened 2018-06-26
Post pr review

iherman:
Went through the full schema.org mapping & co, identified a number of small issues. The ToC and the 'title' items needed some more major re-write.

@mattgarrish I hope it will not create issues with merge...

### DIFF
--- a/experiments/w3c_rec/full_version.html
+++ b/experiments/w3c_rec/full_version.html
@@ -13,7 +13,7 @@
                 "@language"      : "en-US"
             }
         ],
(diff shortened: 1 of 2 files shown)


iherman:
I was tempted to include an editing for #245 (which simply looks like a bug to me), but I decided to leave this to @mattgarrish ...

mattgarrish:
Looks fine to me. I'll merge and then see what it does to my local edits. :)

(I have added some text to deal with id on scripts.)

PR #252, opened 2018-06-28
Updated the moby dick manifest examples

iherman:
@dauwhe @BigBlueHat 

I updated the files in the repo (experiments/separate_manifest) to be (hopefully...) in line with the latest draft. Actually, I also made a 'simpler' version, making use of the shorthand possibilities.

@mattgarrish it may be a good idea to pull the simple version, as well as the simple version of the W3C rec example, as full examples in the appendix...


iherman:
Actually, @mattgarrish, unless it becomes too long, pulling in the audiobook example of @HadrienGardeur might also be a good idea.

mattgarrish:
I wonder if we should just link out to the raw pages in github at this stage, and then consider whether to pull them in once everything stabilizes?

(It might be good to link out, regardless, as it gives greater flexibility post-publishing to make any necessary changes to the examples if we later discover anything missing or in need of improvement.)

mattgarrish:
Although I suppose we could do a data-include with respec, too. Hm...

Issue #251, opened 2018-06-27
table of contents in resources?


mattgarrish:
Is this example correct: https://w3c.github.io/wpub/#table-of-contents ?

I'm a bit confused how the resource list can refer to a fragment of a resource. Does the file have to be listed again without the fragment to be valid?

mattgarrish:
Not questioning whether you get a resource, but what are the implications when referencing a fragment. What if the resource containing the table of contents is already in the default reading order, for example, does this also override the statement that says to only list resources that aren't in the default reading order? So the same document reference does occur in two places?

I'm not sure I see what benefit we gain over just having a separate property that references the toc. It seems like just another piece of potential confusion to add to all these lists and how to use them.

iherman:
Hm. It was one of the changes we agreed to a few weeks ago to _remove_ the separate property references the ToC and, instead, fold the ToC into the general way of handling external links: #234 and we [resolved](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2018/2018-06-25-pwg.html#resolution2) to merge the corresponding PR.

There were good arguments in #234 for a unified approach, I do not think we should go "back" on that.

A possibility to move forward is to say the following: 

1. Indeed, there should be no duplication of resources in the combined set of `readingOrder`, `resources`, and `links`.
2. if a resource is also labeled as `contents` in the corresponding `rel` value (we were careful enough to specify that `rel` is an _array_ of possible values, thanks @HadrienGardeur :-) then the user agent should _find_ an element with `role=doc-toc` in that resource, much like it does with the primary entry page in case no reference to a toc is given in the manifest.

I actually believe this makes much cleaner.

mattgarrish:
> I actually believe this makes much cleaner.

That's better in terms of getting rid of the fragment identifier, yes, but doesn't that also mean you have to look in either the default reading order or the resource list to find the table of contents? I don't have a problem with that, but we should specify it, as well.

iherman:
You are right, that should be specified as well.

iherman:
@BigBlueHat can we try to avoid coming back to issues that have been closed? This issue was discussed and a resolution in favor of the current draft has been accepted, see:

https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2018/2018-06-25-pwg.html#resolution2

Actually, this issue should also be closed in light of that resolution.

mattgarrish:
Closing this since I opened it and I'm fine with the changes and resolution.

PR #254, opened 2018-06-28
Table of contents clarifications

mattgarrish:
This PR addresses the concern I raised in #251 by removing the fragment identifier for table of contents, as recommended by @iherman.

In trying to rewrite the section, I also noticed that we were mixing a lot of toc construction requirements into the property definition, when all the property does is identify the resource containing the toc. To fix this, I've created a new subsection under "Creation" to detail how to make the table of contents, and rewritten the infoset to specify how to obtain the link.

Of note is that it disallows a fragment identifier on PublicationLinks, and also explicitly disallows a resource from being listed in both the resource list and default reading order, or listed two or more times within either of those lists.

Comments welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/254.html" title="Last updated on Jun 29, 2018, 12:47 PM GMT (7e2c2ea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/254/2582155...7e2c2ea.html" title="Last updated on Jun 29, 2018, 12:47 PM GMT (7e2c2ea)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -444,7 +444,7 @@
 					this case the link element MUST use a relative URL to refer to the manifest. </p>
 
 				<p> The manifest itself MUST NOT include a reference to itself, i.e., the reference to the manifest MUST
-					NOT appear as part of the <a href="#wp-resource-categorization-properties"></a>. </p>
+					NOT appear as part of the <a href="#resource-categorization-properties"></a>. </p>
 
 				<p>There are no restrictions on a Web Publication beyond this requirement. The Web Publication MAY
 					include references to resources of any media type, both in the <a>default reading order</a> and as
@@ -625,7 +625,7 @@
 					<h2><code>PublicationLink</code> Definition</h2>
 
 					<p id="publication-link-def">This specification defines a new type for links called
-								<dfn><code>PublicationLink</code></dfn>. It consists of the following properties:</p>
+							<code>PublicationLink</code>. It consists of the following properties:</p>
 
 					<table class="zebra">
 						<thead>
--- a/index.html
+++ b/index.html
@@ -643,7 +643,8 @@
 									<code>url</code>
 								</td>
 								<td>Location of the resource.</td>
-								<td>A URL with no or empty fragment identifier&#160;[[!url]].</td>
+								<td>A URL&#160;[[!url]]. Refer to the property definitions that accept this type for
+									additional restrictions.</td>
 								<td>
 									<a href="https://schema.org/url"><code>url</code></a>
 								</td>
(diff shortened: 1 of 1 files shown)


iherman:
@mattgarrish, on the second item (if you agree) I believe the cleanest approach is:

- in 4.3.1. remove the restriction on the value of `url` referring to the fragment identifier
- both in 4.7.1.1. and 4.7.2.1. add a restriction explicitly in the text, saying something like "URL-s identifying resources MUST NOT contain fragment strings".

I do not want to make the change without your agreement, though.

mattgarrish:
> but I do not think we can restrict values in links (ie, extra links) which are largely URL-s that the authors do not really control.

Funny, I had the same thought last night as I was falling asleep (thinking of wpub is better than counting sheep!).

I was thinking that we make the fragment conditional in the PublicationLink definition, though, to avoid having to repeat it elsewhere. For example:

> A URL [[!url]]. The URL MAY contain a fragment identifier only when used with ???

(Do we call that section "links" now, so we have to say "links in the links property"?)

llemeurfr:
Such a solution seems fine to me. Links (and this is why this name is adequate) are not always complete resources.
-L
> but I do not think we can restrict values in links (ie, extra links) which are largely URL-s that the authors do not really control.
> 
> Funny, I had the same thought last night as I was falling asleep (thinking of wpub is better than counting sheep!).
> 
> I was thinking that we make the fragment conditional in the PublicationLink definition, though, to avoid having to repeat it elsewhere. For example:
> 
> A URL [[!url]]. The URL MAY contain a fragment identifier only when used with ???
> 
> (Do we call that section "links" now, so we have to say "links in the links property"?)

iherman:
WP as sheep:-)

There are some issues that may lead to other 'links' (e.g., links of images, stuff like that), where similar restrictions may apply. Also, we may end up using a schema.org object later to replace it (pending some issues to be discussed with DanBri & Co).

All this to say that It may be more future safe to add that remark to the individual items rather than the overall definition of PublicationLink imho.

HadrienGardeur:
+1 to what @iherman suggested: these restrictions apply to the collection (`resources` and `readingOrder`) rather than the link model.

mattgarrish:
> may be more future safe to add that remark to the individual items rather than the overall definition of PublicationLink

It's just a painful approach as the links property is the lone anomaly.

But I'm not going to get worked up over it. It might just also be good to make clear in the publicationlink definition that people should refer to each property definition for possible additional restrictions.

HadrienGardeur:
@mattgarrish from a JSON-LD perspective, `links` is the only place where fragments make sense, but for the WebIDL we'll also have fragments in `toc` as well (and potentially other lists/navigation, such as page list or landmarks).

iherman:
@mattgarrish 

> But I'm not going to get worked up over it. It might just also be good to make clear in the publicationLink definition that people should refer to each property definition for possible additional restrictions.

Big +1 to that.

Will you do these changes?

mattgarrish:
Yup, just getting the little one off and then will push a fix.

mattgarrish:
I've pushed a change to spell out the restriction in each section. (I also tweaked the "Extra resources" section to use "Links", but will return to that later.)

iherman:
@mattgarrish this should be merged, as far as I am concerned...

PR #257, opened 2018-07-03
Updated the extension section

### LINKED ISSUE #162: Linking to an external metadata record
We have consensus that the manifest should not deal with specialized metadata and rely instead on external metadata records.

What we haven't discussed (yet) is how we actually link to these external metadata records.

It's worth pointing out that this issue is strictly about external metadata records **about the publication**, and not external metadata records about a specific resource contained in the publication.

iherman:
See #162. If accepted, this fix #162

The example files as well as the wp.js file have also been updated.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/257.html" title="Last updated on Jul 5, 2018, 10:36 AM GMT (0a8414c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/257/4529234...0a8414c.html" title="Last updated on Jul 5, 2018, 10:36 AM GMT (0a8414c)">Diff</a>


mattgarrish:
There might be too much structure still right now. I'm fine with flattening, but I think we should drop 4.8.1 as a subsection as the manifest sections are sucked under an infoset heading. It might be simpler as:

> 4.8 Extensibility
> 4.8.1 Linked Records
> 4.8.2 Additional Properties in the Manifest

I think we might want to say more in the section on additional properties, too, like recommend that authors only include properties from public vocabularies like schema.org and dublin core, for example.

We might also want to say something about proprietary extensions, as technically there is nothing that will stop people from adding non-descriptive properties if they choose. Do we want people to use prefixes and/or a registry, like EPUB does in places, and WAM also does for extending?

iherman:
@mattgarrish, can you have a look again?

mattgarrish:
I'm still a bit confused by the first sentence of 4.8.1. Is this subsection only about descriptive properties, as in:

> Additional descriptive terms, beyond those defined in this document, can be included in the manifest.

What about things that aren't considered descriptive, though? If we're addressing those later, then I'm otherwise fine with it.

iherman:
I think the problem is to discuss 'linked metadata records': we do not even use the term 'metadata'.

I think that the second paragraph in the section covers informative and structural properties via the same mechanism. I am not sure how to formulate it right, but that seems to be what 'linked metadata records' mean, in fact...

I have to call it a day, so if you want to reformulate that section, go ahead, I won't touch it for the coming 12 hours:-)

mattgarrish:
Yes, they're probably better called linked records, and those would tend to only cover descriptive properties. That's fine as far as it goes.

Adding "descriptive" to the section on additional properties in manifest suggests that other properties are excluded, similar to how the first sentence focused on additional terms beyond those in the linked descriptive section.

I read the second sentence about proprietary additions in that context, even though I agree it would be the same extensibility regardless of the property type. I'd rather just take "descriptive" out, to be honest.

I pushed a couple of minor changes to this effect. See if it still makes sense.

iherman:
@mattgarrish I like what you did. I think we have successfully beaten this to death:-) I would propose to merge it.

Cc: @llemeurfr

llemeurfr:
The overall wording is fine! I'll go into details that may bother new readers: 

a) I still see the term _descriptive_ in the first list item of the introduction (the text is not completely de-descriptive-fied). Shall we remove it?

b) The sentence  "User Agents MAY support additional properties but MUST NOT include unrecognized properties in the infoset ." is still fuzzy to me. The infoset is defined by the spec, a US cannot do anything to it. But a UA which supports additional properties will logically include them in the memory structure (object) processed from the manifest (a sort of internal infoset). So what does the sentence means?

c) The fragment "as the use of native formats" opens the question of what is a "native format" ?. Shouldn't we use "standard metadata formats" instead, which "simplify processing by user agents"?

d) In 4.8.1, we use "properties", "terms", "public vocabularies". But "terms" is a very wide token (e.g. dcterms is about property names, not values). And using "public vocabularies" is fuzzy (is it about properties or values)?  I suggest using "properties" from **public schemes** (I won't propose ontology) and "values" from **controlled vocabularies**. I'm confident the such vocab. could be used consistently over the spec. 

A side note: it's funny to see XML namespaces chime in in JSON via compact URIs. Not sure the JSON rebellion will like it. 

e) The examples 29 and 30 are not indented properly in the Diff view, but it may be only because it's a Diff.

f) Is the @type = PublicationLink really needed in the sample 31, as we are including the wp-context which should define it?

g) I wonder why linked records would be set in the resource list. Can you point me to the rationale of this choice?

llemeurfr:
@iherman,

Re. the @type, we'll have to check this.
Re. the links vs resource, I'm now convinced by the packaging argumentation. Ok.

iherman:
Re the `@type`: here is what I tried in the playground:

```
{
	"@context" : {
		"test" : {
			"@id"   : "http://example.org/test",
			"@type" : "http://example.org/Class"
		},
		"prop" : {
			"@id" : "https://example/org/classProperty"
		}
	},
	"@id" : "https://example.com/subject",
	"test" : [
          "something",
          {
		"prop" : "something here"
	}]
}
```

and here is what I get in N-Triples:

```
<https://example.com/subject> <http://example.org/test> "something"^^<http://example.org/Class> .
<https://example.com/subject> <http://example.org/test> _:b0 .
_:b0 <https://example/org/classProperty> "something here" .
```

Ie, the class type (as opposed to the datatype) is ignored:-(

mattgarrish:
Does this address your concerns @llemeurfr , @iherman ?

Would be good to close this one soon.

iherman:
I hate brining in a new problem at this point but... I'm beginning to feel uncomfortable with the MUST on using `describeby`. For example, if I use the external linking approach (as I should) to link to a licence or copyright, also using `describedby` sounds a bit irrelevant.

What about:

"the `rel` value of the `PublicationLink` SHOULD also include the `describedby` (IANA) identifier unless there exist a closely relevant identifier defined by IANA or by other organizations;"

Actually, in this terminology, even using MUST for SHOULD may be o.k., but that would raise the question of what 'closely relevant' means...

That detail aside, I think we should ship it.

llemeurfr:
@iherman I understand your concern, as the original idea is to tag links to descriptive metadata with rel=describedby but leave some freedom to tag other links with other rel values.  
But I don't see your proposal explaining that clearly. What about:
 "the rel value of the PublicationLink SHOULD include a relevant identifier defined by IANA or by other organizations; if the link record contains descriptive metadata it MUST include the describedby (IANA) identifier;"

Apart from that I'm happy will the changes, @mattgarrish. I particularly like the sample using a 'dc' extension. That's clear.  Thanks.

iherman:
I am fine with your text, laurent.

iherman:
@mattgarrish I have committed a text with the changes above. If you are fine with that text, you should have the honor of merging this PR!

PR #258, opened 2018-07-03
Added the contributor as a creator.

### LINKED ISSUE #256: Support for a generic contributor element
[Our current draft](https://w3c.github.io/wpub/#wp-creator-manifest) has a fairly exhaustive list of contributors but it's missing a generic contributor element: http://schema.org/contributor

We'll need that element for contributors that do not fit one of the roles currently in our draft or for which there's no equivalent in schema.org.

iherman:
See #256. If accepted and merge fix #256

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/258.html" title="Last updated on Jul 5, 2018, 2:44 PM GMT (24c6333)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/258/072cee5...24c6333.html" title="Last updated on Jul 5, 2018, 2:44 PM GMT (24c6333)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -949,12 +949,7 @@
 					<section id="canonical-identifier-infoset">
 						<h5>Infoset Requirements</h5>
 
-						<p>The canonical identifier MUST be an <a>address</a> or a value that allows a one-to-one
-							mapping to an address (e.g., a <abbr title="Digital Object Identifier"
-							>DOI</abbr>&#160;[[doi]] that can be resolved to a <abbr title="Uniform Resource Locator"
-								>URL</abbr>&#160;[[url]] via a <abbr title="Digital Object Identifier">DOI</abbr>
-							resolver). If a user agent cannot resolve the canonical identifier to an address, it SHOULD
-							ignore the property.</p>
+						<p>The canonical identifier MUST be an <a>address</a>.</p>
 
 						<p>If a Web Publication is hosted at more than one address, the canonical identifier allows a
 							user agent to identify the shared relationship between the versions and to determine which
@@ -974,10 +969,15 @@
 							conventions of the identifier scheme used and the degree of control over assignment of
 							identifiers.</p>
 
-						<p class="note">If the canonical identifier is a <abbr title="Uniform Resource Locator"
-								>URL</abbr>, it can be used as the target of a "canonical" link&#160;[[rfc6596]] (e.g.,
-							a <code>link</code> element&#160;[[html]] whose <code>rel</code> attribute has the value
-								<code>canonical</code> or an <abbr title="Hypertext Transfer Protocol">HTTP</abbr>
+						<p>If a URL is not provided in the manifest, or the value is an invalid URL, the Web Publication
+							does not have a canonical identifier. In this case, a user agent MUST NOT make any
+							assumptions about the relationship between the current Web Publication and any other
+							possible versions of it (i.e., treat each instance as unique).</p>
+
+						<p class="note">The canonical identifier can be used as the target of a "canonical"
+							link&#160;[[rfc6596]] (e.g., a <code>link</code> element&#160;[[html]] whose
+								<code>rel</code> attribute has the value <code>canonical</code> or an <abbr
+								title="Hypertext Transfer Protocol">HTTP</abbr>
 							<code>Link</code> header field&#160;[[rfc5988]] similarly identified).</p>
 
 						<p class="issue" data-number="58">Is a canonical identifier necessary to call out explicitly in
@@ -1000,39 +1000,25 @@
 							<tbody>
 								<tr>
 									<td>
-										<code>id</code>/ <code>@id</code>
+										<code>id</code> / <code>@id</code>
 									</td>
 									<td>Preferred version of the Web Publication.</td>
 									<td>A URL [[!url]].</td>
-									<td> (None. <code>id</code> maps to the <a
-											href="https://www.w3.org/TR/json-ld/#node-objects"><code>@id</code></a>
-										keyword [[!json-ld]].) </td>
-								</tr>
-								<tr>
-									<td>
-										<code>identifier</code>
-									</td>
-									<td>Preferred version of the Web Publication.</td>
-									<td>A textual string that represents a unique identifier, or a <a
-											href="https://schema.org/PropertyValue"><code>PropertyValue</code> type</a>
-										[[!schema.org]].</td>
-									<td>
-										<a href="https://schema.org/identifier"><code>identifier</code></a>
-									</td>
+									<td>(None)</td>
 								</tr>
 							</tbody>
 						</table>
 
-						<p>The use of <code>id</code> is RECOMMENDED when the identifier is a URL.</p>
+						<p>The manifest MUST NOT contain both <code>id</code> and <code>@id</code> declarations &#8212;
+							only one is permitted.</p>
 
 						<p class="note">The <code>id</code> term and the <code>@id</code> keyword can be used
 							interchangeably: the former is mapped onto the second by the schema.org context.</p>
 
-						<p>There are also sub-properties for <code>identifier</code> that can be used on their own
-							right, like <a href="https://schema.org/isbn"><code>isbn</code></a>, <a
-								href="https://schema.org/issn"><code>issn</code></a>, or <a
-								href="https://schema.org/serialNumber"><code>serialNumber</code></a>. See also the <a
-								href="https://schema.org/identifier">Schema.org</a> description for further details.</p>
+						<p>The specification of a canonical identifier does not preclude the use of the <a
+								href="https://schema.org/identifier"><code>identifier</code>
+							property</a>&#160;[[!schema.org]] or its subtypes to specify other identifiers for a Web
+							Publication.</p>
 
 						<pre class="example" title="Example for setting both the canonical identifier as URL and the address of the same document">
 {
--- a/index.html
+++ b/index.html
@@ -1110,6 +1110,17 @@
 								</tr>
 								<tr>
 									<td>
+										<code>contributor</code>
+									</td>
+									<td>Contributor whose role does not fit to one of the other roles in this table.</td>
+									<td>One or more <a href="https://schema.org/Person"><code>Person</code></a> or <a
+											href="https://schema.org/Organization"><code>Organization</code></a>.</td>
+									<td>
+										<a href="https://schema.org/contributor"><code>contributor</code></a>
+									</td>
+								</tr>
+								<tr>
+									<td>
 										<code>creator</code>
 									</td>
 									<td>The creator of the publication.</td>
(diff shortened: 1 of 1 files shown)


mattgarrish:
There might be too much structure still right now. I'm fine with flattening, but I think we should drop 4.8.1 as a subsection as the manifest sections are sucked under an infoset heading. It might be simpler as:

> 4.8 Extensibility
> 4.8.1 Linked Records
> 4.8.2 Additional Properties in the Manifest

I think we might want to say more in the section on additional properties, too, like recommend that authors only include properties from public vocabularies like schema.org and dublin core, for example.

We might also want to say something about proprietary extensions, as technically there is nothing that will stop people from adding non-descriptive properties if they choose. Do we want people to use prefixes and/or a registry, like EPUB does in places, and WAM also does for extending?

mattgarrish:
Is it concerning to anyone but me that we're meshing creators and contributors all together under a single heading of creators? From a display priority perspective, for example, are all these creators treated equally?

laudrain:
@mattgarrish In ONIX we only have Contributor for all sort of ... contributors. And it's ok!
Obviously, we also have a ContributorRole and a SequenceNumber.
As soon as we find the same in the WP Infoset, it'ok to have only one keyword creator.
It is the case in [4.4.4 Creator](https://w3c.github.io/wpub/#creator) for the role, the sequence is still in question with schema.org.

iherman:
@mattgarrish unavoidably the name of the term will carry some "semantics", that UA will use as they see fit. Otherwise we do not really have a choice: we cannot ignore the proliferation of those schema.org terms that all have their value.

In this sense, the "creator" is more a generic entry in the infoset, and the original name should be kept somehow. I guess this means a change in the final WebIDL that should use the original manifest name as some sort of a description of the creator(s). @HadrienGardeur ?

llemeurfr:
What bothers me in the commit is that the generic _creator_ and _contributor_ values are lost in a set of specializations of this very notion. When reading the spec, I'd prefer seing the generic terms set before their specialization, with a statement like "Precise roles in the creative process can be expressed using one ont the following properties instead of the generic creator and contributor properties".

mattgarrish:
The problem with mixing is that it's not required to continue to distinguish these roles at the infoset level. Retaining the specialized roles is only a "should", in which case an author is indistinguishable from an editor. Is there a reason why this isn't a "must"? I might be more comfortable with that change.

But, again, with all these specializations, and with them all being compiled into a single generic infoset property, are there rules on how to go about this? Display precedence is always an issue with creators, so what do authors and user agents need to take into consideration? If it's just "do your best" we're going to have inconsistencies in ordering and what even gets displayed. Should the person who checked the markup be listed right there with the authors, for example, since they're all creators?

HadrienGardeur:
@iherman I'll have to update the WebIDL as well.

There are two possible solutions for the WebIDL:

- group all contributors in the same list and have a "role" element
- align with JSON and add many different lists of contributors

iherman:
I believe that the cleanest approach is your first option, @HadrienGardeur. Actually, I believe that should be defined in a similar way in the infoset as well. We have a "creator", which is also characterized by a "role". The mapping on Schema is therefore a means to define the specific "roles", as defined by the schema.org vocabulary.

HadrienGardeur:
The first option is the EPUB approach, but I don't think that this will work well with JSON-LD + schema.org.

llemeurfr:
Sorry, this discussion is difficult to understand. Shall we use different property names; a single creator property + role as sub-property; a creator and a contributor property + role? 

json samples of what you advocate would be welcome.

iherman:
@HadrienGardeur @llemeurfr : nothing should change in the JSON. The only change would be in the processing of the Manifest: when seeing, say, the "editor" key in the manifest, this would be mapped, in the infoset into the "creator" infoset item, with a role set to "editor".

Having, in the infoset, all the various schema.org terms listed separately is also possible of course (your second option, @HadrienGardeur) is also possible of course, but it would make the infoset as well as the WebIDL pretty messy and large; it would also be difficult to keep it open ended, whereas one would expect new creator roles to pop up at schema.org.

But maybe @mattgarrish can find some neat editorial trick:-)

HadrienGardeur:
@iherman OK thanks for the clarification.

The only issue I have with adding a role element to a single list of contributors in the WebIDL is related to ordering.
How do we order things in a single list that were initially broken down into multiple elements in JSON?

mattgarrish:
Do we want to change "Creators" to a categorization of properties instead of a specific property of the infoset? We have the same problem with "accessibility" where we have one property right now that says it is a listing of a variety of other properties.

It would make the WebIDL longer, but if the property definitions are all the same we could limit making a mess of the properties section by not dealing with each individually.

Only thought I have on keeping them unique within the infoset.

iherman:
@mattgarrish I knew you would have some editorial trick up our sleeve:-) But the analogy with accessibility is a good point, which may make it worth doing it after all.

The WebIDL will become awfully long, but maybe that is simply the nature of the beast.

iherman:
@HadrienGardeur it is your turn to provide some clarification, I am not sure I understand:

> How do we order things in a single list that were initially broken down into multiple elements in JSON?

Do you simply refer to the fact that schema.org does not order things right now? This is an open issue with schema.org/DanBri, and we take care of it right now in the json ld context...

mattgarrish:
I think it's the same question I had about you compile all the different types of creators up into a single "creator" property? Distinguishing them is helpful, but you still have to create a master list from them. Keeping them distinct pushes this problem further down the stack to how does a user agent present them, but it's still lingering in the background.

mattgarrish:
Could we perhaps call "readBy" the "narrator" in the infoset? If we want harmony with schema.org I'll shut up, but every time I see it as a classification it bothers me.

mattgarrish:
I'll do some mods to the PR, since I offered up the categorization idea (and look again at accessibility and identifiers), but will leave the webidl alone while I do.

iherman:
Let us leave 'readBy' as it is. There a number of schema.org terms I personally do not like, but we should not go down the bikeshedding line:-)

mattgarrish:
I can live with that rationale, even if I've never met a professional readby before... ;)

mattgarrish:
I pushed some changes for accessibility/creators, but I'm confused what to do about canonical identifiers. Shouldn't there be a restriction to only one, but if we restrict the canonical identifier to only one, how is that done without preventing a publication from having multiple identifiers?

Which of identifier or its subtypes gets treated as canonical, in other words?

mattgarrish:
Sorry, push of the changes to the wrong PR has been fixed now.

iherman:
Minor things

* "Creator" -> "Creators" in 4.2. Similarly, "Accessibility report" -> "Accessibility"
* With the new structure, I would move the reference to the http://www.w3.org/wiki/WebSchemas/Accessibility to the infoset description subsection in Accessibility. That page gives a general description rather than "just" schema.org terms
*

iherman:
@mattgarrish 

> but I'm confused what to do about canonical identifiers. Shouldn't there be a restriction to only one, but if we restrict the canonical identifier to only one, how is that done without preventing a publication from having multiple identifiers?

Sigh. Yes, this is getting messy, too.

Clearly, there may be only one `@id`.

But the text currently, sort of, equals the usage of `@id` and `identifier`. And that may be the wrong approach. An alternative: there should be _one_ canonical ID which, if used, MUST be a URL, _and_ there may be a number of additional identifiers, whose content and values depend on various communities. This is a change on the infoset, and it can then be translated to schema. 

I do not think the standard should even begin to enumerate the various types of identifiers on the infoset level, only that they exist, there can be several ones, and there SHOULD be a unique one per category (whatever that means...). The "subproperty" in the schema.org mapping is then there for specific ones.

(One of the questions with schema.org is how to define new subproperties, eg, to add a DOI. I think we can leave that open for now in the document, with a note that this is subject to further discussions.)

llemeurfr:
@iherman I would love to see approved the notion of one @id (as a URL, which must be globally unambiguous -> same id => same publication) and a set of alternative identifiers with a scheme and a value (which may be any string).

mattgarrish:
I agree with @llemeurfr. If we're going to have one canonical identifier, we need a clear way of providing it that doesn't require the user agent to determine which of several possibilities might be canonical.

The section should probably only identify @id for the canonical identifier and say that the specification of the canonical identifier does not affect the ability to use identifier and its subtypes to specify other identifiers for the specification, or something to that effect.

mattgarrish:
I've made some modifications to canonical identifier if you want to have a look. I don't believe we're preventing DOIs by requiring an address, since a DOI can be expressed as a URL. In the case of Web Publications, you'd just be required to use the URL form. (Expecting the user agent to resolve non-URLs is probably asking too much, regardless.)

I tried to added more depth than "ignore the property" when an address is not provided, but feel free to critique.

mattgarrish:
> "hosted at more than one address" shouldn't that be "hosted at more than one addresses"

No, "one address" is singular so it's fine. "hosted at multiple addresses" would be plural. But either works. Plural probably sounds a little cleaner.

I don't like the note about id/@id after setting the restriction. It basically explains the restriction, so would make more sense merged into the same paragraph.

See if the latest commits strikes a better balance.

llemeurfr:
Just to be sure that I understand well (and others too), we've got:

- the canonical identifier (id) = an address (URL) which is stable in time, i.e. it must not change if the WP evolves or is moved between different locations; it may be a WP address, **or not**; 
- identifier(s) (from schema.org) = any string value (may be a URL or not), which scheme is expressed via the property name (a sub-property of identifier, e.g. isbn) 
- the WP address = an address (URL), that cannot be used as an identifier because a WP may be duplicated at different locations. As a consequence, a WP may have several WP addresses. It references the entry page of the WP (HTML). In this entry page one finds a reference to the WP manifest or the WP manifest itself. 
- the address of the WP manifest = an address (URL), that exists only if the manifest is not embedded in the entry page, and that has no specific description in the spec. 

Am I right? 

I still feel unsure about the relationship between the canonical id and the WP address. The following snippets in bold, from 4.4.3.1, show some contradictions:

"If a Web Publication is hosted at more than one address, the canonical identifier allows a user agent to identify the shared relationship between the versions __and to determine which of the available addresses is primary__.

The canonical identifier is also intended to provide a measure of permanence above and beyond the Web Publication's address . Even if a Web Publication is permanently relocated to a new address, for example, __the canonical identifier will provide a way of locating the new location__ (e.g., a DOI registry could be updated with the new URL , or a redirect could be added to the URL of the canonical identifier).

When assigned, the canonical identifier needs to be unique to one and only one Web Publication, __independent of its address(es)__."

I would prefer to keep the last sentence = canonical id and WP address(es) are independent, period.

HadrienGardeur:
I think that the discussion about identifiers should be handled elsewhere (issue tracker, not in a PR), where it'll have more visibility.

iherman:
@mattgarrish I am fine with your latest changes on the text.

I agree with @HadrienGardeur: this PR actually started on the missing `contributor`, and it somehow got diverted into the identifier. I would propose to merge both this and the other (#257) PR asap, and we may then have a cleaner way of discussing the final text on the identifiers.

mattgarrish:
Yes, let's merge and continue talking identifiers in a new issue. @llemeurfr if you want to copy your concerns above to a new issue, that would help.

Issue #259, opened 2018-07-03
reading progression - conflict in spec


TzviyaSiegman:
There is a conflict in the section on reading progression direction (https://w3c.github.io/wpub/#reading-progression-direction)
section 4.4.8.1 says "The default value is auto."
section 4.4.8.2 says "If this value is not set, its default value is ltr."

@mattgarrish do you recall what we concluded?

mattgarrish:
Don't recall a discussion, but why wouldn't "auto" be the default? That's the value that says the author didn't define anything, or has no preference, so use whatever is natural for the reader.

iherman:
The note at the end of https://www.w3.org/TR/html52/dom.html#the-dir-attribute says 'ltr'. We should go with that...

mattgarrish:
It also technically doesn't have a default value:

> The attribute has no invalid value default and no missing value default.

You only get to ltr through the algorithm, which we don't have an equivalent of. In fact, our statement about how to compute auto is flawed:

> direction is determined from the value of the property.

How can you determine the direction from the word "auto" or from the absence of the property altogether, so no value?

Also, "auto" should never be a value of the infoset, should it? It would only be valid in the manifest, as the infoset would have to have the computed ltr or rtl value. In which case, do we need "auto" at all, or only an algorithm to specify ltr/rtl when not set?

mattgarrish:
Or is "auto" trying to say that you have to calculate a value every time you parse an infoset value? That would make a little more sense, but I still wonder whether it really belongs as a possibility in the infoset.

iherman:
I am fine ditching 'auto' altogether...

mattgarrish:
Ya, I can't wrap my head around it, as the infoset contains the computed values so the direction has to be one or the other by that stage. It's only within the manifest that it could be unspecified.

But should we consider a better algorithm for computing? Like html dir, if it is not specified, ltr is assumed. If auto is specified at the manifest level, maybe the title is inspected first for directionality?

Maybe I'll make a PR with some thoughts in this direction.

PR #262, opened 2018-07-03
base direction cleanup

mattgarrish:
Not sure this is ready for a commit, but offering up for comments to continue the discussion in #259.

Another oddity I found is that there are steps to calculate the language for the language, ending in undetermined, but the manifest says the default is always English. Should the manifest define defaults, or should it be the infoset? Might be good to get consensus so we don't end up mixing.

More to the point, I've removed the manifest default language for now. I recall we already went through not using "en" as a default earlier in the revision and that's how we got to "und", but if it's changing back again we have to swap out the other specification.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1268,7 +1268,6 @@
 						<ul>
 							<li><code>ltr</code>: left-to-right;</li>
 							<li><code>rtl</code>: right-to-left;</li>
-							<li><code>auto</code>: direction is determined from the value of the property.</li>
 						</ul>
 
 						<p>When specified, these properties are used as defaults for textual values in the <abbr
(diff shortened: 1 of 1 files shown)


iherman:
- I do not remember the discussion leading to "und" but _I think_ that is the official value if no language tag is specified in HTML. We should indeed do the same
- I am not sure about the handling of direction "auto" in our case, though. I presume if no direction is specified for the titles, the result will be... "ltr'? Also, what happens if there are several titles, each with different directionality values?

I am tempted to ban "auto" altogether. However, the situation becomes worse due to the inability to set the direction (as opposed to HTML) in JSON-LD. The only solution that is becoming viable is to rely on the UTF markers only, much like the activity stream people have done, see https://www.w3.org/TR/activitystreams-core/#h-biditext. If that is the way we go, this fundametnally changes the relevant portions of the spec, because directionality gets out of the manifest altogether...

At the moment this is flagged as completely open in the spec, so maybe we should leave this part open for now, and contact Richard & Co right away when the current draft is published.

mattgarrish:
I'm happy to get rid of "auto" entirely, but I don't remember why we added it so limited myself to removing it from the compiled infoset for the first pass.

How about we strip it and see what reaction it gets rather than try to patch around it?

iherman:
It may be as simple as something we took over from HTML because it was there...

The problem I have, rather, is as follows. If we define, as we do, the explicit global/local dir value in the infoset, we should be able to extract it from the manifest. On the other hand, if we go down the line of "embedding" the direction value into the text value itself via the UTF special characters (like in the activity streams) than I begin to wonder whether it makes sense to have those infoset values explicitly in the first place. We _may_ remove them altogether in favor of some very general statement making it clear that we expect the textual values to take care of directionality.  (I still have to look at the activity stream document in more details.)

This is something that we should really get discussed with the I18N experts asap, rather than coming up with something of our own and then hit their review later...

mattgarrish:
Right, I don't have any answer to that problem. All I'm suggesting for now is we reduce the complexity of having "auto", which even the HTML spec advises authors not to use.

mattgarrish:
Effectively, all we say is that if the base direction isn't specified, or set to an invalid value, use ltr. Auto and the rules for determining go.

mattgarrish:
See the new preview.

iherman:
The only minor thing is that pushing the reference to issue 220 right to the top may be good. At the moment there are sentences in the text:

> The value of the base direction MUST be "ltr" or "rtl", as appropriate.

but there is nothing in the text that makes it clear _how_ those values should be set. Pushing the reference up in the text may make it clear that this is a known missing thing.

Otherwise I would propose to merge it and put it aside for now.

mattgarrish:
Okay, I've moved the note to immediately after where base direction is first defined at the start of the section.

PR #266, opened 2018-07-09
Fixes #260. Clarifying cover infoset requirements.

### LINKED ISSUE #260: Cover infoset requirements is unclear
In the WP draft, [Structural Properties: Cover: Infoset Requirements](https://w3c.github.io/wpub/#cover-infoset) is unclear:

> User agents SHOULD NOT use the cover image as the sole means of selecting or accessing Web Publications. A user agent SHOULD use the Web Publication's title and creators as text alternatives for such interfaces.

deborahgu:
***
<a href="https://pr-preview.s3.amazonaws.com/deborahgu/wpub/pull/266.html" title="Last updated on Jul 9, 2018, 4:07 PM GMT (51622ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/266/f5ac2a7...deborahgu:51622ab.html" title="Last updated on Jul 9, 2018, 4:07 PM GMT (51622ab)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1789,9 +1789,10 @@
 						<p>The <abbr title="information set"><a>infoset</a></abbr> SHOULD include a reference to a cover
 							image.</p>
 
-						<p>User agents SHOULD NOT use the cover image as the sole means of selecting or accessing Web
-							Publications. A user agent SHOULD use the Web Publication's <a href="#title">title</a> and
-								<a href="#creators">creators</a> as text alternatives for such interfaces.</p>
+						<p>A user agent SHOULD provide the Web Publication's <a href="#title">title</a> and
+						    <a href="#creators">creators</a>, in textual form, as access points for a Web
+                            Publication. User agents SHOULD NOT rely on the cover image, with or without alternative
+                            text, as the sole means of selecting or accessing Web Publications.</p>
 
 						<p>More than one cover image MAY be referenced from the infoset to provide alternative sizes and
 							resolutions for different device screens.</p>


mattgarrish:
I'm still uneasy with the first sentence in that it normatively indicates how a user agent is expected to provide access in a bookshelf, etc. That isn't specifically related to the cover or its compilation from the manifest.

It's sort of light foray into UAAG territory. We might want to reconsider whether these points need saying, or if there's a way of referencing UAAG, where relevant, in the user agent section.

deborahgu:
Yeah, I was thinking about that.  I was mostly just addressing the lack of clarity in my rewriting, and not addressing the overall point of the initial statement, which did address UAs.  Because in a straight-up description of the infoset, no part of this statement belongs, but it _does_ have to exist somewhere.

I mean, it could be something like:

> A cover image is not considered to be a sufficient access point for a publication. A textually-rendered title and creators are necessary and sufficient access points for a publication.  

It's not normative, but at least it's explicitly stated.

I'm beginning to think that we need  "WP Accessibility" to parallel [EPUB Accessibility](http://www.idpf.org/epub/a11y/accessibility.html). If we remove this, it won't be the first accessibility statement we've removed from the draft because it didn't belong in the definition of WP, per se, but that nevertheless needs to be documented as a minimum for WP accessibility.

The reality is that, just as with Epub Accessibility, we're going to have some requirements or recommendations that aren't encapsulated by WCAG or UAAG, and we need to record them _somewhere_.

mattgarrish:
I'm going to close this as we've addressed it, I believe, through #273. There is a "should" to include a title and description and an example of their use.

And just because it's odd to come off sounding like I'm advocating against accessibility, I think there are relevant statements being raised here that we should take up when we address how features are implemented. It's just that at the infoset level that we should only be concerned with "affording" features, like accessible interfaces, and I believe we have that.

Issue #261, opened 2018-07-03
use of "cover image"


TzviyaSiegman:
https://w3c.github.io/wpub/#cover

Some in the group expressed interest in allowing the cover to be something other than an image. The cover section is mostly written with the assumption that the content is an image.

deborahgu:
In the conversation about this, didn't we agree that all of the non-image covers people could thing of, eg [ibooks](https://images.duckduckgo.com/iu/?u=http%3A%2F%2Ftidbits.com%2Fresources%2F2011-12%2FiBooks-public-domain-covers.png&f=1) or  [scholarly paper covers](https://images.duckduckgo.com/iu/?u=http%3A%2F%2Fwww.cs.utah.edu%2F~knolla%2Fthesis_cover.png&f=1), were still generated thumbnails? I mean, it would be cool if they didn't have to be, except we didn't here from anyone (that I recall) that there's actually a demand for rendering a thing called a "cover" as non-images.

TzviyaSiegman:
@deborahgu I don't remember. Maybe @atyposh recalls?

llemeurfr:
The cover may be a raster or vector image, it is still an image. 

I’m curious to hear from those who’d like some other kind of content, which kind they would see fit the notion of cover: video? Audio? Text?

laudrain:
@TzviyaSiegman and @deborahgu  in the Toronto F2F I find:
> Charles LaPierre: Cover’s don’t have to be images.

Non-image covers could still be captured as an image and included as such in WP.

HadrienGardeur:
A cover could also be HTML + `<canvas>`.

The New Yorker is a good example, they often have "interactive covers" that would work nicely with a `<canvas>` element.

avneeshsingh:
In Toronto face to face we decided to allow use of HTML for cover image. It can have an image with alttext and also animation or video with equivalent text. It is a good approach for accessibility.
The pending issue was regarding use of aria role. The aria role coverimage can be applied to only images as per DPUB ARIA module, how can we extend its usage to other HTML elements.

danielweck:
Whilst I agree that "rich" covers offer valuable design opportunities (from basic vector images to full-blown animated and interactive visuals), there is also tremendous value in static raster images (for example, to render as dumb "thumbnail" covers in a catalog / library view).

In that light, would the distinction between "cover page" and "cover image" be useful?

TzviyaSiegman:
[...]

PR #273, opened 2018-07-23
allow any type of cover

mattgarrish:
Based on the latest comment from @TzviyaSiegman in #261, here's a rewrite of the cover section to allow any type of resource. I included a "must" that when there are multiple covers specified they be unique on at least one property. I also added "should" for title and description, and expanded the examples.

I removed any mention of creating a cover or how to use it, as it's not clear what the user agent would create or when. Given the open-ended nature, this will have to be covered in any sections that make use of the cover, or left to user agents to provide on their own.

Comments welcome if I missed anything from today's discussion.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/273.html" title="Last updated on Jul 24, 2018, 1:29 PM GMT (5620de1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/273/ff607cd...5620de1.html" title="Last updated on Jul 24, 2018, 1:29 PM GMT (5620de1)">Diff</a>

### DIFF
--- a/common/js/wp.js
+++ b/common/js/wp.js
@@ -79,33 +79,33 @@ function create_wp(config) {
         "accessModeSufficient" : ["textual"],
 
         "resources"            : [{
-            "@type"       : "PublicationLink",
-            "url"         : "https://www.w3.org/StyleSheets/TR/2016/logos/W3C",
-            "fileFormat"  : "image/svg+xml",
-            "description" : "W3C Logo"           
+            "@type"           : "PublicationLink",
+            "url"             : "https://www.w3.org/StyleSheets/TR/2016/logos/W3C",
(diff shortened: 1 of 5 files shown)


iherman:
One thing, maybe: we did say at the call that we include the more "permissive" version in the draft but asking the community to comment on this issue because there was no 100% consensus. Maybe a reference to an issue and an editorial note on this would be good.

iherman:
(If merged) this fixes #261 fixed #260 and closes #266

mattgarrish:
I've added a reference to issue 261 since it already contains the discussion to date, so we should leave it open after merging.

PR #274, opened 2018-07-24
Added a section clarifying what the base is for manifest

iherman:
Just added a minor section stating the obvious: the base URL is either the manifest's URL when the manifest is a separate file or the base URL of the enclosing HTML, when it is embedded.

(This is currently not specified in JSON-LD 1.0, because the usage of `script` is non-normative, and the correct standardization of that feature in JSON-LD 1.1 might come way later. If JSON-1.1 becomes a standard before WP, this section may become superfluous...)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/274.html" title="Last updated on Jul 24, 2018, 3:09 PM GMT (1a4a883)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/274/007d7fc...1a4a883.html" title="Last updated on Jul 24, 2018, 3:09 PM GMT (1a4a883)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -459,6 +459,24 @@
 				</div>
 			</section>
 
+			<section>
+				<h3>Resolving URLs in Manifests</h3>
+
+				<p>
+					Manifests may contain <a href="https://url.spec.whatwg.org/#relative-url-string">relative URL strings</a>. To resolve such URLs into <a href="https://url.spec.whatwg.org/#absolute-url-string">absolute URL strings</a>, a <a href="https://url.spec.whatwg.org/#concept-base-url">base URL</a> is necessary (see&nbsp;[[!url]]). This base URL for the manifest is:
+				</p>
+
+				<ul>
+					<li>in the case of an <a href="#manifest-embedding">embedded Manifest</a> this is the <a href="https://www.w3.org/TR/html/infrastructure.html#document-base-url">document base URL</a>&nbsp;[[!html]] of the <a>primary entry page</a> of the Web Publication;</li>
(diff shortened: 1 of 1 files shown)


mattgarrish:
Only question I have is whether this might be a better fit under "3.2 Creating a Manifest"? (e.g., "3.2.4 Resolving URLs")

iherman:
I was not sure either. I let you decide on that...

mattgarrish:
My preference would be to move it there. It just feels a bit detached where it is, as the section has gone on to embedding and linking and then resources before returning to what feels like a key piece of info about creating manifests (although I realize this is also important for processing manifests, not just creating them, but creators need to understand this).

mattgarrish:
See if you're okay with what I've done. I changed the title to try to make it fit the section better and did some minor tweaking. There's no great place for this, as I don't like talking about embedding and linking contexts before we reach them. One place or the other it will be awkward, so not something I want to dither on any more. :)

iherman:
@mattgarrish minor issue: I am not sure what is exactly the fate of `xml:base` in (X)HTML. Ie, strictly speaking, `<base>` may not be the _only_ way of setting the base URL, that is why I had an "e.g." in there. (Even if all other means are pretty theoretical...)

mattgarrish:
Ah, right, forgot about that. I've put it into a parenthetical and mentioned/linked xml:base.

iherman:
@mattgarrish will you do the honor of merging? Or should I? :-)

mattgarrish:
Me! Me! Me! :)

PR #277, opened 2018-07-25
Pre publ editing

### LINKED ISSUE #259: reading progression - conflict in spec

iherman:
Lots of minor editing; the most important are:

- added a paragraph to the status section on where we are
- fix #259 (this went down the cracks, removing `auto` from the reading order)
- the description on how to access the table of content seemed to be wrong, rewrote this (@mattgarrish check this!)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/277.html" title="Last updated on Jul 25, 2018, 2:39 PM GMT (49c1b58)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/277/f33f89c...49c1b58.html" title="Last updated on Jul 25, 2018, 2:39 PM GMT (49c1b58)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1568,11 +1568,10 @@
 
 						<ul>
 							<li><code>ltr</code>: left-to-right;</li>
-							<li><code>rtl</code>: right-to-left;</li>
-							<li><code>auto</code>: the user agent chooses the direction.</li>
+							<li><code>rtl</code>: right-to-left.</li>
 						</ul>
 
-						<p>The default value is <code>auto</code>.</p>
+						<p>The default value is <code>ltr</code>.</p>
 
 						<p>This infoset item has <em>no effect</em> on the rendering of the individual primary
 							resources; it is only relevant for the progression direction from one resource to the
@@ -1601,7 +1600,7 @@
 										<code>readingProgression</code>
 									</td>
 									<td>Reading direction from one resource to the other.</td>
-									<td><code>ltr</code>, <code>rtl</code>, or <code>auto</code></td>
+									<td><code>ltr</code> or <code>rtl</code></td>
 									<td>(None)</td>
 								</tr>
 							</tbody>
@@ -2160,6 +2159,7 @@
 							</blockquote>
 							<p>This was not decided on the Toronto F2F, and is still open.</p>
 						</div>
+						<p class=issue data-number=205></p>
 					</section>
 
 					<section id="resource-list-manifest">
--- a/index.html
+++ b/index.html
@@ -2219,7 +2219,7 @@
 									<li>
 										If a resource in either the <a href="#default-reading-order">default reading order</a>
 										or <a href="#resource-list">resource-list</a> is identified with a <code>rel</code>
-										value including <code>contents</code>&#160;[[!iana-link-relations]], the corresponding <code>url</code> value identifies the table of content resource. If there is more than one such resource, the user agent MUST use the first one (beginning with the <a href="#default-reading-order">default reading order</a>).
+										value including <code>contents</code>&#160;[[!iana-link-relations]], the corresponding <code>url</code> value identifies the table of content resource.
 									</li> 
 									<li>
 										Otherwise, the <a>primary entry page</a> is the table of content resource.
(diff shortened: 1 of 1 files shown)


iherman:
@mattgarrish I was also wondering whether we shouldn't change the order slightly, and move Resource Categorization properties (4.7) before 4.5? Entries like cover, toc, privacy policy refer to the resource list or links, which is defined later. That may be a bit disturbing to first time users...

mattgarrish:
A few minor things:

- in 4.4.4. Creators -- "There are represented by" s/There/Creators/
- 4.5.1.1. isn't this an editor's note? Seems to indicate things we still need to work out.
- 4.6.2.1 generally fine with the reformulation, but wondering if there's some equivalent to document order for discovering the "contents" resource. "first one" doesn't imply any direction for checking, which should seem obvious that you start at the beginning of the array, but if epub has proven anything...
- should reference be to whatwg dom or dom 4?

iherman:
> in 4.4.4. Creators -- "There are represented by" s/There/Creators/

+1

> 4.5.1.1. isn't this an editor's note? Seems to indicate things we still need to work out.

Well... not sure. Whoever produces the accessibility report can do it in different format: JSON, XML, HTML, or a yet-to-be-invented-super-format. Isn't it up to the author which one to be put into the manifest (possibly using content negotiations to choose among formats)? Why is it something the manifest should do and, mainly, how?

> 4.6.2.1 generally fine with the reformulation, but wondering if there's some equivalent to document order for discovering the "contents" resource. "first one" doesn't imply any direction for checking, which should seem obvious that you start at the beginning of the array, but if epub has proven anything...

(Furthermore, the resource-list is not even order dependent.) But I am not sure what to replace it with. We could say this is undefined or be even tougher and to say that if you put several TOC references then all of them are invalid...

To be honest, I am not sure what to replace this with. Leave it to implementers?

> should reference be to whatwg dom or dom 4?

At this point I'd leave it to respec (or, more exactly, to specref) to make the choice.

iherman:
@mattgarrish @TzviyaSiegman I took care of your comments and I have also reordered the sections for the manifest items. (https://github.com/w3c/wpub/pull/277#issuecomment-407752688). Merge?

mattgarrish:
> Isn't it up to the author which one to be put into the manifest

Hm, I think this note has gone astray, as it's not about the accessibility record, which you're right is another place that could contain the metadata. This was originally about whether accessibility metadata belongs locked up in the manifest where it may not be available to users (until there is search support, at least), or whether it should be included in a more transparent location.

At the very least, it's not a fully formed note, as the language is unclear about how the infoset would get extended:

> Machine-readable accessibility metadata may be recommended in whatever format is used to externalize publication metadata ... Depending how this externalizing is done

It could be made into a note, but it needs rewriting to take out the vagueness. (It also needs to be moved back to the accessibility section, of course.)

TzviyaSiegman:
I think we got a little lost in versions, but to @mattgarrish 's point, maybe we should add a note that indicates that the infoset is extensible, and future version will clarify details?

iherman:
@mattgarrish will you take care of editing that one? To be honest, I am not sure what to do with it...

mattgarrish:
> (Furthermore, the resource-list is not even order dependent.)

Ya, yuck, that doesn't help things.

Maybe there should be a specific authoring instruction that only one resource can be identified as the table of contents. That would make the presence of two an error handling situation that we can defer to user agents.

iherman:
Something like:

"If identifying the Table of Content is ambiguous (e.g., several table of content resources are identified, or several more than one elements with a role value doc-toc is found within the table of content resource), the user agent MAY choose among them. This specification does not mandate how this choice is made. The user agent might:

* choose the first table of content resource by some heuristics
* choose the first table of content element in document tree order.
"

This would then replace the second sentence of the 1.1 item, as well as the paragraph just after the bullet list.

(I copied the same style as what we have for language choice. Actually, we may have to discuss at some point, and look at each item, what type of conformance problems may occur and what should be the reaction of the user agent...)

Issue #284, opened 2018-07-30
Mixture of Person and Organization?


iherman:
The current manifest says for, e.g., "publisher" that it is an array of Persons _or_ Organizations. It does not say whether the array should be homogeneous, ie, whether contains objects of identical types. The current manifest suggests that it is not strictly homogeneous, but it is something that may have to be decided. 

Is there a use case where the list of, say, publishers, can include some persons mixed with some organizations?

laudrain:
In Publishers I don't think so, the list would be only Organizations: co-publishers.
But In Creator, Contributor, there may be a mix of Persons AND Organizations.

iherman:
Thanks @laudrain.

I must admit I would prefer not to have different constraints for different categories. Ie, if, for some categories like creators, a mixed list is o.k., then I would prefer to keep it open for all categories...

mattgarrish:
> I must admit I would prefer not to have different constraints for different categories.

This seems to come from schema.org itself, but it doesn't appear to be worth the bother of enforcing. There are always edge cases, and all excluding one class does is force you to report an organization as a person or a person as an organization.

iherman:
@mattgarrish @laudrain 

What we say is that we leave it as it is: all these arrays might be "non-homogeneous". I am fine with that.

Although the normative text is fine, do you think it is worth putting an editorial note somewhere making this clear @mattgarrish ?

mattgarrish:
Ya, I've been scratching my head trying to think of a clear way to say it in the tables so we don't have to note across properties:

- One or more Person and/or Organization
- One or more of: Person, Organization
- A mix of one or more Person or Organization

Preferences? Other ideas?

laudrain:
> One  or more Person and/or Organization 

has my preference.

mattgarrish:
> is grammatically correct... does that work for our purposes

As a normative statement, yes, but right now we have tables that define the properties and indicate their required values ("One or More Person or Organization") in a specific column. If we can find an equally short replacement for these, then the information will be apparent even on lazy reading. (And who doesn't like it when you can be lazy and still get the information you need? :)

tcole3:
Regarding the original question, simultaneous publication can result in multiple publishers, and though very uncommon currently, recording a person as publisher used to be more common and could become more than an edge case pattern again. Also library catalog rules conflate distributor and publisher, which further increases chance of a mixed array for publisher. But I agree much more likely in creator role. 

[...]

PR #285, opened 2018-07-30
WebIDL update

iherman:
My first attempt to bring the WebIDL part in sync with the current manifest. Some of the significant changes (some of those may be subject to further discussions!)

- The previous IDL referred to "title" as an array of strings, which was correct. I have made a slight update on the manifest definition emphasizing the fact that the 'name' value may be an array, too.
- The previous IDL included a "children" tag, which did not have a direct counterpart in the manifest. I believe the main reason for its presence was to handle hierarchical TOC-s. However, the manifest as of today extracts/contains a _link_ to a TOC, itself encoded in HTML, and not the list of resources themselves. Of course, a user agent may want to extract those (or may choose to simply display the HTML) but I am not sure it is a matter of the IDL to have this.
- I was wondering about having a single "Contributors" dictionary item for all the authors, editors, etc., with an extra flag providing the categorization, but I thought spelling things out, ie, reflecting what is in the manifest, is cleaner...

There are also some other, minor changes, mostly editorial.

See also #284

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/285.html" title="Last updated on Aug 1, 2018, 12:01 PM GMT (bcf7927)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/285/9b7c71e...bcf7927.html" title="Last updated on Aug 1, 2018, 12:01 PM GMT (bcf7927)">Diff</a>

### DIFF
--- a/webidl/manifest.webidl
+++ b/webidl/manifest.webidl
@@ -40,7 +40,7 @@ dictionary WebPublicationManifest {
              sequence<PublicationLink>          links;
 
              DOMString                          accessibilityReport;
-             DOMString                           privacyPolicy;
+             DOMString                          privacyPolicy;
              sequence<DOMString>                cover;
              DOMString                          toc;
 };
--- a/webidl/manifest.webidl
+++ b/webidl/manifest.webidl
@@ -39,8 +39,8 @@ dictionary WebPublicationManifest {
              sequence<PublicationLink>          resources;
              sequence<PublicationLink>          links;
 
-             DOMString                          accessibilityReport;
-             DOMString                          privacyPolicy;
-             sequence<DOMString>                cover;
+             PublicationLink                    accessibilityReport;
+             PublicationLink                    privacyPolicy;
+             sequence<PublicationLink>          cover;
              DOMString                          toc;
 };
--- a/webidl/manifest.webidl
+++ b/webidl/manifest.webidl
@@ -2,13 +2,13 @@ dictionary WebPublicationManifest {
     required DOMString                          url;
     required DOMString                          type;
 
-             DOMString                          accessMode;
-             DOMString                          accessModeSufficient;
+             sequence<DOMString>                accessMode;
+             sequence<DOMString>                accessModeSufficient;
              DOMString                          accessibilityAPI;
-             DOMString                          accessibilityControl;
-             DOMString                          accessibilityFeature;
-             DOMString                          accessibilityHazard;
-             DOMString                          accessibilitySummary;
+             sequence<DOMString>                accessibilityControl;
+             sequence<DOMString>                accessibilityFeature;
+             sequence<DOMString>                accessibilityHazard;
+             LocalizableString                  accessibilitySummary;
 
              DOMString                          id;
 
--- a/webidl/manifest.webidl
+++ b/webidl/manifest.webidl
@@ -40,7 +40,7 @@ dictionary WebPublicationManifest {
              sequence<PublicationLink>          links;
 
              DOMString                          accessibilityReport;
-             DOMSring                           privacyPolicy;
+             DOMString                           privacyPolicy;
              sequence<DOMString>                cover;
              DOMString                          toc;
 };
--- a/webidl/manifest.webidl
+++ b/webidl/manifest.webidl
@@ -1,15 +1,46 @@
 dictionary WebPublicationManifest {
-    required DOMString                         url;
-             DOMString                         lang;
-             TextDirection                     direction = "auto";
-             TextDirection                     readingProgression = "auto";
-             sequence<LocalizableString>       name;
-             DOMString                         id;
-             sequence<Contributor>             authors;
-             DOMString                         dateModified;
-             DOMString                         datePublished;
-             sequence<PublicationLink>         links;
-             sequence<PublicationLink>         readingOrder;
-             sequence<PublicationLink>         resources;
-             sequence<PublicationLink>         toc;
+    required DOMString                          url;
+    required DOMString                          type;
+
+             DOMString                          accessMode;
+             DOMString                          accessModeSufficient;
+             DOMString                          accessibilityAPI;
+             DOMString                          accessibilityControl;
+             DOMString                          accessibilityFeature;
+             DOMString                          accessibilityHazard;
+             DOMString                          accessibilitySummary;
+
+             DOMString                          id;
+
+             sequence<Person>                   artist;
+             sequence<(Person or Organization)> author;
+             sequence<Person>                   colorist;
+             sequence<(Person or Organization)> contributor;
+             sequence<(Person or Organization)> creator;
+             sequence<Person>                   editor;
+             sequence<Person>                   illustrator;
+             sequence<Person>                   letterer;
+             sequence<Person>                   penciler;
+             sequence<(Person or Organization)> publisher;
+             sequence<Person>                   readby;
+             sequence<(Person or Organization)> translator;
+
+             DOMString                          inLanguage;
+             TextDirection                      inDirection;
+
+             DOMString                          dateModified;
+             DOMString                          datePublished;
+
+             ProgressionDirection               readingProgression = "ltr";
+
+             sequence<LocalizableString>        name;
+
+             sequence<PublicationLink>          readingOrder;
+             sequence<PublicationLink>          resources;
+             sequence<PublicationLink>          links;
+
+             DOMString                          accessibilityReport;
+             DOMSring                           privacyPolicy;
+             sequence<DOMString>                cover;
+             DOMString                          toc;
 };
--- a/webidl/organization.webidl
+++ b/webidl/organization.webidl
@@ -1,0 +1,7 @@
+
+dictionary Organization {
+    required sequence<LocalizableString> name;
+             DOMString                   id;
+             DOMString                   url;
+};
+
--- a/webidl/person.webidl
+++ b/webidl/person.webidl
@@ -1,0 +1,6 @@
+
+dictionary Person {
+    required sequence<LocalizableString> name;
+             DOMString                   id;
+             DOMString                   url;
+};
--- a/webidl/publication_link.webidl
+++ b/webidl/publication_link.webidl
@@ -1,7 +1,7 @@
 dictionary PublicationLink {
-    required DOMString                  url;
-             DOMString                  encodingFormat;
-             DOMString                  name;
-             sequence<DOMString>        rel;
-             sequence<PublicationLink>  children;
+    required DOMString           url;
+             DOMString           encodingFormat;
+             DOMString           name;
+             DOMString           description;
+             sequence<DOMString> rel;
 };
--- a/webidl/publication_link.webidl
+++ b/webidl/publication_link.webidl
@@ -1,7 +1,7 @@
 dictionary PublicationLink {
     required DOMString           url;
              DOMString           encodingFormat;
-             DOMString           name;
-             DOMString           description;
+             LocalizableString   name;
+             LocalizableString   description;
              sequence<DOMString> rel;
 };
--- a/experiments/w3c_rec/full_version.html
+++ b/experiments/w3c_rec/full_version.html
@@ -65,6 +65,11 @@
                 "url"   : "https://www.w3.org"
             }
         }],
+         "publisher" : {
+            "@type" : "Organization",
+            "name" : "World Wide Web Consortium",
+            "@id"  : "https://www.w3.org/"
+        },
         "datePublished"         : "2015-12-17",
         "dateModified"          : "2015-12-17",
         "resources"             : [
--- a/experiments/w3c_rec/simple_version.html
+++ b/experiments/w3c_rec/simple_version.html
@@ -26,7 +26,12 @@
                 "name"          : "Ivan Herman"
             }
         ],
-        "datePublished"         : "2015-12-17",
+         "publisher" : {
+            "@type" : "Organization",
+            "name" : "World Wide Web Consortium",
+            "@id"  : "https://www.w3.org/"
+        },
+       "datePublished"         : "2015-12-17",
         "resources"             : [
             "datatypes.html",
             "datatypes.svg",
--- a/index.html
+++ b/index.html
@@ -1215,7 +1215,7 @@
     },{
         "@type" : "Person",
         "name" : "Ivan Herman",
-        "url"  : "https://www.w3.org/People/Ivan/"
+        "@id"  : "https://www.w3.org/People/Ivan/"
     }],
     "editor"    : [{
         "@type" : "Person",
@@ -1227,7 +1227,7 @@
     "publisher" : {
         "@type" : "Organization",
         "name" : "World Wide Web Consortium",
-        "url"  : "https://www.w3.org/"
+        "@id"  : "https://www.w3.org/"
     }
     &#8230;
 }
--- a/index.html
+++ b/index.html
@@ -2914,9 +2914,6 @@
 		<section id="webidl" class="appendix">
 			<h2>WebIDL</h2>
 
-			<p class="ednote">The WebIDL is not yet fully aligned with the infoset. This section will be updated in a
-				future release.</p>
-
 			<section id="webidl-intro" class="informative">
 				<h3>Introduction</h3>
 
@@ -2935,7 +2932,7 @@
 
 				<pre class="idl" data-include="webidl/manifest.webidl" data-include-format="text"></pre>
 
-				<p>The <a>WebPublicationManifest</a> has the following members:</p>
+				<p>The <a><code>WebPublicationManifest</code></a> has the following members:</p>
 
 				<dl data-dfn-for="WebPublicationManifest">
 					<dt>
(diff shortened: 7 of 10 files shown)


llemeurfr:
@iherman re. title, an array of arrays is a serious modification of the model. What makes you think that this is needed? 

The [use cases I know of](https://github.com/readium/webpub-manifest/tree/master/contexts/default) are: 
- alternative languages
- title + subtitle
- alternate wording for search. 

title is not the same as e.g. contributor, where each item is a different entity.

iherman:
> @iherman re. title, an array of arrays is a serious modification of the model. What makes you think that this is needed?

I mean "array of strings", not array of arrays... To be clear, it should be possible to say:

```
"name" : ["War and Peace", "Guerre et Paix"]
```

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
+
+             ProgressionDirection               readingProgression = "ltr";
+
+             sequence<LocalizableString>        name;
+
+             sequence<PublicationLink>          readingOrder;
+             sequence<PublicationLink>          resources;
+             sequence<PublicationLink>          links;
+
+             DOMString                          accessibilityReport;
```
HadrienGardeur: Why do we need a specific string for that? It should be already covered by `links`/`resources`.

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
+             DOMString                           privacyPolicy;
```
HadrienGardeur: Same comment, should already be covered by `links`/`resources`.

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
+             sequence<DOMString>                cover;
```
HadrienGardeur: Same comment, should already be covered by `links`/`resources`.

webidl/publication_link.webidl
```diff
@@ -1,7 +1,7 @@
 dictionary PublicationLink {
-    required DOMString                  url;
-             DOMString                  encodingFormat;
-             DOMString                  name;
-             sequence<DOMString>        rel;
-             sequence<PublicationLink>  children;
```
HadrienGardeur: A UA will need to parse the HTML containing the ToC to extract it. I think that there's no reason to use a separate structure for handling the ToC when the WebIDL can play that role as well.

webidl/publication_link.webidl
```diff
@@ -1,7 +1,7 @@
...
+    required DOMString           url;
+             DOMString           encodingFormat;
+             DOMString           name;
```
HadrienGardeur: This means that `name` won't be localizable.

webidl/publication_link.webidl
```diff
@@ -1,7 +1,7 @@
...
+             DOMString           description;
```
HadrienGardeur: This means that `description` won't be localizable.

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
-             sequence<PublicationLink>         resources;
-             sequence<PublicationLink>         toc;
+    required DOMString                          url;
+    required DOMString                          type;
+
+             DOMString                          accessMode;
+             DOMString                          accessModeSufficient;
+             DOMString                          accessibilityAPI;
+             DOMString                          accessibilityControl;
+             DOMString                          accessibilityFeature;
```
HadrienGardeur: This is expressed using an array of strings on the example available at schema.org: https://schema.org/accessibilityFeature

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
-             sequence<PublicationLink>         readingOrder;
-             sequence<PublicationLink>         resources;
-             sequence<PublicationLink>         toc;
+    required DOMString                          url;
+    required DOMString                          type;
+
+             DOMString                          accessMode;
+             DOMString                          accessModeSufficient;
+             DOMString                          accessibilityAPI;
+             DOMString                          accessibilityControl;
```
HadrienGardeur: This is expressed using an array of strings on the example available at schema.org: https://schema.org/accessibilityFeature

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
+             DOMString                          accessibilityFeature;
+             DOMString                          accessibilityHazard;
```
HadrienGardeur: This is expressed using an array of strings on the example available at schema.org: https://schema.org/accessibilityFeature

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
-             DOMString                         dateModified;
-             DOMString                         datePublished;
-             sequence<PublicationLink>         links;
-             sequence<PublicationLink>         readingOrder;
-             sequence<PublicationLink>         resources;
-             sequence<PublicationLink>         toc;
+    required DOMString                          url;
+    required DOMString                          type;
+
+             DOMString                          accessMode;
```
HadrienGardeur: This is expressed using an array of strings on the example available at schema.org: https://schema.org/accessMode

webidl/manifest.webidl
```diff
@@ -1,15 +1,46 @@
...
+             DOMString                          accessModeSufficient;
```
HadrienGardeur: This is expressed using an array of strings on the example available at schema.org: https://schema.org/accessMode

webidl/organization.webidl
```diff
@@ -0,0 +1,7 @@
+
+dictionary Organization {
+    required sequence<LocalizableString> name;
+             DOMString                   id;
```
HadrienGardeur: We need an example for this in the spec.

webidl/person.webidl
```diff
@@ -0,0 +1,6 @@
+
+dictionary Person {
+    required sequence<LocalizableString> name;
+             DOMString                   id;
```
HadrienGardeur: We need an example for this in the spec.

HadrienGardeur:
```
"name" : ["War and Peace", "Guerre et Paix"]
```

@iherman I find this example quite confusing and would recommend that we limit `name` to:

- string
- object
- array of object

Whenever an object is used, both `@value` and `@language` should be required.

llemeurfr:
@iherman we are touching there again on a very important issue, that cannot be discussed in this PR, I agree: are we creating a sort of "semantic metadata framework", in which authors can generate flexible data given their intimate knowledge of JSON-LD? or are we creating a interchange format as a JSON structure, that authors and UA must closely follow, validated by a JSON schema, but which "by design" is JSON-LD compliant?

I'll open a specific issue for that.

webidl/publication_link.webidl
iherman: Changed in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/publication_link.webidl
iherman: Changed in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/manifest.webidl
iherman: Changed in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/manifest.webidl
iherman: Changed in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/manifest.webidl
iherman: Changed in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/manifest.webidl
iherman: Changed in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/manifest.webidl
iherman: Changed in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/organization.webidl
iherman: Added in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

webidl/person.webidl
iherman: Added in https://github.com/w3c/wpub/pull/285/commits/aefe40f27a55ebf37d4272c8d887d32d5bcd2bf3

HadrienGardeur:
@iherman I just saw your latest commit. For localizable strings, I think that we actually need a sequence of them since a string can be localized in multiple languages.

iherman:
@HadrienGardeur 

I have made the changes except for the introduction of accessibilityReport, privacyPolicy, cover, and toc.

You are absolutely right that those are extracted from resources or links, and that is how the manifest is defined. However, as the infoset calls this out explicitly, I felt that the WebIDL should reflect, as much as possible, the infoset, too, and that is why I included them there. I do not feel very strongly about it, but I think that it makes the relationship to the infoset more explicit.

See what others feel about this...

HadrienGardeur:
@iherman even if we keep separate IDL elements for `cover`, `privacyPolicy` and `accessibilityReport`, these can't be represented as a string or a sequence of strings.

Since we would be duplicating information from `links`, `resources` or `readingOrder`, we should at least be consistent with how they're expressed and use a `PublicationLink` or a sequence of `PublicationLink` for those.

I think that the `toc` is unique, in the sense that the UA will have to fetch and parse HTML to properly populate this info. I know that you have your doubts about that, but this is definitely something that every UA will need to do and I don't see any good reason why it can't be included in the same IDL.

iherman:
> Since we would be duplicating information from links, resources or readingOrder, we should at least be consistent with how they're expressed and use a PublicationLink or a sequence of PublicationLink for those.

You are right. That is a bug; this is necessary for things like using the description of a cover image as an alternate text. I will take care of this soon.

The problem with toc: we then have to spec exactly how the structure of the toc should be expressed in HTML, mainly when it comes to hierarchical toc-s.

HadrienGardeur:
> The problem with toc: we then have to spec exactly how the structure of the toc should be expressed in HTML, mainly when it comes to hierarchical toc-s.

@iherman this is IMO unavoidable.

llemeurfr:
@iherman, if not, UAs won't be able to do anything interesting with the HTML ToC. 
The alternative we proposed was a predefined machine readable json ToC but it was dismissed by the group.

TzviyaSiegman:
We could consider restricting the way that the HTML ToC can ve written as is done in EPUB. See https://w3c.github.io/publ-epub-revision/epub32/spec/epub-packages.html#sec-package-nav-def.

iherman:
@TzviyaSiegman @HadrienGardeur that works for me, if the community is (and has been) fine with it).

@mattgarrish will have a pleasure reproducing his own text in the new draft;-)

Back to the original item fir this PR, I guess the WebIDL would have to be modified by

1. re-introducing the child reference in the PublicationLink
2. the TOC is a sequence of Publication links.

mattgarrish:
Funny, I was just reading the nav restrictions again while reviewing 3.2. There have to be ways we can loosen the restrictions and still have parseable data (e.g., ol or ul as the list type, even if semantic purity is lost, treating text outside of ```a``` tags as decorative and ignorable, etc.).

TzviyaSiegman:
> There have to be ways we can loosen the restrictions and still have parseable data (e.g., ol or ul as the list type, even if semantic purity is lost, treating text outside of a tags as decorative and ignorable, etc.).

This seems like a good idea to me. Provide direction to machines, not restrictions to humans.

dauwhe:
If we start profiling HTML in web publications, we will likely alienate the browser community as well as massively confuse authors.

danielweck:
@iherman
> "The problem with toc: we then have to spec exactly how the structure of the toc should be expressed in HTML, mainly when it comes to hierarchical toc-s."

@HadrienGardeur 
> "this is IMO unavoidable."

@llemeurfr 
> "if not, UAs won't be able to do anything interesting with the HTML ToC. The alternative we proposed was a predefined machine readable json ToC but it was dismissed by the group."

@TzviyaSiegman
> "We could consider restricting the way that the HTML ToC can ve written as is done in EPUB."

We've been down that rabbit hole before :)

I do not know how widespread this criticism of EPUB3 Navigation Document is, but here is what I heard on numerous occasions:

The fact that the NavDoc is a "microdata" model restriction on what would otherwise be totally arbitrary HTML markup has forced (some) content creators to:

(1) exclude the NavDoc from the spine reading order / list of render-able documents (i.e. linear or non-linear resources), effectively treating the mandatory unique NavDoc HTML only as a machine-readable data source.
(2) author a separate HTML document (unfortunately not marked with any sort of "NavDoc" metadata, so not reliably discoverable by a reading system) in order to gain full markup flexibility and to provide fancy layout / styling (e.g. SVG instead of HTML lists).

This simply reflects the fact that (some) content creators prefer ; at the cost of some degree of duplication ; a clear distinction between:

1) optimized, machine-readable data structure designed to be ingested by a reading system / user agent, with the expectation that it will be rendered in an arbitrary, application-specific fashion.
2) authored web document (custom structure and semantics, desired styling), with the expectation that it will be faithfully rendered by a web browser (or an web browser engine / webview component inside a reading system).

This dichotomy could be expressed in Web Publications by having a separate machine-readable Table Of Contents in the JSON serialization of the manifest infoset, but we excluded this option (if I remember correctly, because of duplication, and also due to HTML offering better string localization than JSON).

I am not suggesting we should revisit this decision, just making observations about how the restricted EPUB3 NavDoc markup syntax was received in the "real world".

TzviyaSiegman:
@danielweck Agreed. I do think we need to look at the way users approach the ToC, not the way authors and publishers believe that users approach the ToC. 

I have not done formal testing, but as a user I always use to machine-rendered ToC because that is what is accessible from all locations in the publication (the always-present hamburger button). I think @mattgarrish's proposal of instructing UAs to disregard unneeded information is a good idea.

mattgarrish:
Is it correct that the table of contents pointed to from the manifest doesn't have to be visible in the page?

This would cover the case of wanting to provide a table of contents but not want it explicitly in the reading order (probably rare, but EPUB doesn't require navigation in spine).

More to the point, though, it could also potentially address Daniel's concern of allowing two tables of contents if whatever parsing algorithm we come up with wouldn't produce a desirable result. I'm hoping we can avoid duplication, but having a plan for the inevitable is also a wise idea.

iherman:
(Purely administrative comment)

It seems that the TOC issue is much more complicated, and I am afraid of a PR staying hanging and becoming problematic to merge if other changes come to the fore. May I suggest:

* In the case the refreshed WebIDL is acceptable for everyone involved _except_ for the TOC item (which is, currently, just a simple pointer to a TOC somewhere else), _let us merge what we have_
* We open a separate issue on the details of the TOC, knowing that a decision may affect the WebIDL later

@HadrienGardeur @mattgarrish @TzviyaSiegman ?

laudrain:
May I recall here the Toronto F2F resolution on TOC:

> TOC should be present
>    - if present, it MUST be an element with doc-toc role pointed to from manifest
>   - <nav> (likely) element in landing page
>   - could elsewhere

In publishing "real world" we need both options Daniel described 

>  a clear distinction between:
> 1. optimized, machine-readable data structure designed to be ingested by a reading system / user agent, with the expectation that it will be rendered in an arbitrary, application-specific fashion.
> 2. authored web document (custom structure and semantics, desired styling), with the expectation that it will be faithfully rendered by a web browser (or an web browser engine / webview component inside a reading system).

If we can have in the same HTML document, a proper structure and styling for TOC AND a data structure for a11y with doc-toc role attribute, it could avoid duplication.
But I have no problem with duplication to achieve these 2 functions and will not give up on any of the 2.

mattgarrish:
> But I have no problem with duplication to achieve these 2 functions and will not give up on any of the 2.

I don't think anything prevents you from creating as many tables of contents as your heart desires. You just point the manifest to whatever one is best for machine processing.

My one concern is that this probably forces a separate document unless both the machine and human versions can exist in the same file without the user encountering both, which is why I wonder if one can be suppressed visually without affecting its parsing (I'm not sure why not, but am just curious).

mattgarrish:
But how did we get on to this in a PR? Where are we in terms of integrating the changes?

iherman:
@mattgarrish 

> But how did we get on to this in a PR? Where are we in terms of integrating the changes?

At the moment, the WebIDL entry for a TOC is simply a reference to the TOC, wherever it is. *If* we had a clear specification of the TOC in terms of HTML, this *could* be replaced by some JSON representation of of the TOC instead.

Hence my proposal to merge this PR as is (unless there are other issues which, at this moment, there aren't any more) and move this discussion to a specific issue.

laudrain:
@mattgarrish Toronto F2F resolution made me hope that it could only one document, particularly in that sentence:
> if present, it MUST be an element with doc-toc role pointed to from manifest

So then I don't think there an issue with this. I agree with @iherman to merge this PR as is

mattgarrish:
Yes, I'm fine with this PR as far as it represents where we are right now. We can re-litigate the toc elsewhere.

PR #293, opened 2018-08-08
Accessibility API should also be a (possible) list values as opposed to a single one

iherman:
The Accessibility API is the only a11y term that had, as a possible value, a single string, as opposed to a comma separated list. If my understanding is correct, the meaning of this attribute is that the publication is compatible with a number of API-s, ie, it is only meaningful if a comma separated set of values are allowed (just like all the other properties). This PR makes this change.

@avneeshsingh @rdeltour @danielweck  is that correct?

(Note that neither the schema.org description more the a11y wiki page is really clear on this...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/293.html" title="Last updated on Aug 9, 2018, 11:54 AM GMT (337c02a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/293/8a8c7d0...337c02a.html" title="Last updated on Aug 9, 2018, 11:54 AM GMT (337c02a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -762,7 +762,7 @@
 									</td>
 									<td> Indicates that the resource is compatible with the referenced accessibility
 										API-s. </td>
-									<td> Comma-separated values. <a
+									<td> Text. <a
 											href="https://www.w3.org/wiki/WebSchemas/Accessibility#accessibilityFeature_in_detail"
 											>Expected values</a>. </td>
 									<td>
--- a/index.html
+++ b/index.html
@@ -761,8 +761,8 @@
 										<code>accessibilityAPI</code>
 									</td>
 									<td> Indicates that the resource is compatible with the referenced accessibility
-										API. </td>
-									<td> Text. <a
+										API-s. </td>
+									<td> Comma-separated values. <a
 											href="https://www.w3.org/wiki/WebSchemas/Accessibility#accessibilityFeature_in_detail"
 											>Expected values</a>. </td>
 									<td>
--- a/webidl/manifest.webidl
+++ b/webidl/manifest.webidl
@@ -4,7 +4,7 @@ dictionary WebPublicationManifest {
 
              sequence<DOMString>                accessMode;
              sequence<DOMString>                accessModeSufficient;
-             DOMString                          accessibilityAPI;
+             sequence<DOMString>                accessibilityAPI;
              sequence<DOMString>                accessibilityControl;
              sequence<DOMString>                accessibilityFeature;
              sequence<DOMString>                accessibilityHazard;


mattgarrish:
Yes, it shouldn't take a single string but should allow an array of values. The only comma-separated string value is accessModeSufficient, but it also takes an array of comma-separated values.

mattgarrish:
The proposed changes don't look right, though. The reason only accessModeSufficient lists comma-separated values is because each instance is a list of one or more modes (technically, it's just a text value, but the text value is expected to be a list of tokens):

```
"accessModeSufficient":  [
   "textual,visual",
   "textual"
]
accessibilityAPI: [
   "ARIA",
    "..."
]
```

HadrienGardeur:
I just fixed the JSON Schema as well.

mattgarrish:
Have a look at the JSON-LD example in schema.org, too: https://schema.org/accessibilityAPI

Each property is basically an array of tokens (excluding the summary, of course). I don't know if there's a better way of saying that accessModeSufficient is an array of comma-separated values, but that's the one distinction it has over the other properties. It really should be an array of arrays, but that's hard to express outside of JSON.

HadrienGardeur:
Will `accessibilityAPI` also allow a single string as well? Just want to make sure that this is covered correctly in the JSON Schema.

HadrienGardeur:
> Have a look at the JSON-LD example in schema.org, too: https://schema.org/accessibilityAPI

The example has a string, not an array of string... I'll modify the schema to allow both for now, but every time we allow such variations, we make things slightly more complex for UAs.

iherman:
@mattgarrish, I am more confused as ever.

- The examples uou refer to on the schema.org side are array of strings. This was my understanding of the situation so far.
- I have also looked at https://www.w3.org/wiki/WebSchemas/Accessibility, which does not mention anything about comma separated lists. Where is that mentioned?
- On the other hand, I see in the example things like `largePrint/CSSEnabled`. The wiki speaks about "transformations" (I am not sure what this means) although `CSSEnabled` is not listed among those.

I think we should suspend the JSON Schema and WebIDL specification on these until we have a clearer picture; can we try to come up with a clear syntax definition?

Also, as a general point (echoing what @HadrienGardeur said) I would be wary of introducing some sort of a micro-syntax within JSON. JSON has a clear syntax on arrays, we already have the practice elsewhere in schema.org (and in the WPM) to use single values as an abbreviation for a single valued array, but it would be good to stay within these...

CC @avneeshsingh

avneeshsingh:
The issue that proposed accessmode and accessmodesufficient is as follows. There was good discussion about syntax to be used:

https://github.com/schemaorg/schemaorg/issues/1100

iherman:
@avneeshsingh although I was part of some intermediate discussions (that I forgot:-) I am looking at the final case in the schema.org documentation. Based on those, the only thing I see from the documentation is that:

- all the accessibility terms, except for `accessibilitySummary` are, potentially, a (JSON) list of terms, like `["textual", "visual"]` for `accessMode`.
- I do not see any example on the schema.org site for structures like the one mentions by @mattgarrish in https://github.com/w3c/wpub/pull/293#issuecomment-411461660 (ie, something like `["textual,visual", "textual"]`. This may be a problem in the schema.org documentation. 
- I do see, however, more complex structures (see the 3rd example in [accessMode](https://schema.org/accessMode)) which we do not even mention in our draft.
- I do see, in the second example of [accessibilityAPI](https://schema.org/accessibilityAPI), that the API may indeed be a list of elements. Which means that my remark starting this discussion holds, the API should be like the others...

iherman:
Looking at the examples, is it so that:

(1) `accessibilitySummary` is a localizable string

(2) `accessModeSufficient` can be:

2. a simple array of strings (with each predefined values)
3. an array of objects of type `ItemList`, with
	1. an `itemListElement` that can be a simple array of strings (with each predefined values)
	2. a `description`, that is a simple text

(3) all other accessibility terms can be a simple array of predefined strings (the list of possible values being dependent on the term itself).

However, how should I then understand the first example of https://schema.org/accessibilityHazard, which has, as an example:

```
"accessibilityFeature": [
    "largePrint/CSSEnabled",
    "highContrast/CSSEnabled",
    "resizeText/CSSEnabled",
    "displayTransformability",
    "longDescription",
    "alternativeText"
  ],
```

That seems to contractict to (2) above...

mattgarrish:
> The examples uou refer to on the schema.org side are array of strings. This was my understanding of the situation so far.

Yes, that's generally the case. The question I have is why did you change API to a comma-separated list? Where did you find that it takes a list?

> I have also looked at https://www.w3.org/wiki/WebSchemas/Accessibility, which does not mention anything about comma separated lists.

It should be more explicit, but the definition of accessModeSufficient says:

> A __list__ of single or combined accessModes that are sufficient to understand all the intellectual content of a resource

You can have one or more sets of sufficient access modes, and each is a comma-separated (assuming there is more than one to list, of course). The wiki description is lacking in this regard, as we continued discussing whether it should be comma or space separated after getting the properties accepted. See the techniques document, which is what we were minting the value for: http://www.idpf.org/epub/a11y/techniques/#meta-002

> can we try to come up with a clear syntax definition?

It's a list of one or more sets of comma-separated values. What alternative definition would you propose? (I'm fine leaving it as "Text", btw, but it loses an important distinction it would be useful to highlight.)

What I read from Hadrien is that the example was not showing an array. Maybe I misunderstood what he's talking about, as we seem to be flipping between API and sufficient. API isn't an array in the schema.org example:

> "accessibilityAPI": "ARIA",

To me, this is perfectly fine, as it's only one value. We're not suggesting that people always use arrays, even if there is only one value, are we?

> I would be wary of introducing some sort of a micro-syntax within JSON

See the discussion that @avneeshsingh referenced. It wasn't my decision or creation. We already went through alternative representations and finally agreed that using a comma-separated list of values was the simplest and most effective solution. I don't think we should change that expectation here.

iherman:
> Yes, that's generally the case. The question I have is why did you change API to a comma-separated list? Where did you find that it takes a list?

Oops, my mistake. I copied from the previous item, I did not realize that was the exception...

> See the discussion that @avneeshsingh referenced. It wasn't my decision or creation. We already went through alternative representations and finally agreed that using a comma-separated list of values was the simplest and most effective solution. 

I do not see any example for that on schema.org! See my summary as I see it on 
https://github.com/w3c/wpub/pull/293#issuecomment-411723178

Or is the schema.org description outdated?

mattgarrish:
> Or is the schema.org description outdated?

Something is wrong with the schema.org definition, because it shows it accepting a text value but the examples do indeed show ListItem.

Like I said above, I'm fine with doing accessModeSufficient as a proper array of lists. If we agree that API is no different than the others, then let's change accessModeSufficient to ListItem and see about changing the schema.org page.

mattgarrish:
Sorry... s/ListItem/ItemList/

mattgarrish:
I'm still confused by the schema.org examples, too. I was looking at example 1, which looks like an array of text values. But example 3 is clearly an array of ItemList values.

Example 1 isn't even correct from the look of it, as there should be a combined textual,visual access mode set.

More things to fix, anyway.

iherman:
@mattgarrish the question is: who fixes what? Is it so that there was no good synchronization between the various participation of the discussion, including Dan? I must admit I do not think _we_ in this group should be the ones coming with a definition; we are supposed to adhere to the schema.org definitions...

iherman:
@mattgarrish I would think that, in practice, we could merge this PR (I have committed a version that removed my mistake, so it is really just the update of the WebIDL) and then there should be some consistent definition somewhere... (maybe a separate issue?)

mattgarrish:
> I must admit I do not think we in this group should be the ones coming with a definition

I speak of "we" in the sense that all, or almost all, of the participants in creating/submitting the metadata are in this group. In that sense, I don't think there's any issue in our submitting patches for the accessModeSufficient definition and examples, but it's not something that this WG as a whole necessarily needs to be involved in.

I now see where I got twisted up between what we had to do for EPUB, the incorrect expected value in the schema.org definition and the different examples. When I saw text expected plus the first example, I thought we abandoned all the more complex models, but I suspect the final submission probably got a copy/paste error from the other definitions.

IMS made the formal submission of these properties, so we should probably coordinate with Madeleine, but again that's something that Avneesh, George, Charles and I (etc.) can do offline.

iherman:
@mattgarrish 

Anything I should do on the WP draft?

HadrienGardeur:
> To me, this is perfectly fine, as it's only one value. We're not suggesting that people always use arrays, even if there is only one value, are we?

Well, that's something we need to decide. Currently, the JSON Schema validates most elements strictly with an array and would reject a string if it's not used for `accessibilityAPI` or `accessibilitySummary`.

iherman:
> To me, this is perfectly fine, as it's only one value. We're not suggesting that people always use arrays, even if there is only one value, are we?

I think we should be consistent, and avoid being restrictive here when we are liberal elsewhere.

mattgarrish:
Let's merge and close this off. We can figure out ItemList for schema.org and then port it over here rather than implement it here first.

My suspicion is people will want to use strings whenever any of these arrays, or any other properties that take arrays of strings only contain one value, but I believe we already have an issue for this.

PR #296, opened 2018-08-09
allow person or organization

mattgarrish:
Per #284, this makes clear that person and/or org are allowed for those properties that accept both. I didn't change any of the properties that only allow person to also allow org, as if we're strictly adhering to schema.org definitions then we shouldn't be adding to them.

Fix #296.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/296.html" title="Last updated on Aug 10, 2018, 5:10 AM GMT (2513682)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/296/c073565...2513682.html" title="Last updated on Aug 10, 2018, 5:10 AM GMT (2513682)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1072,7 +1072,7 @@
 										<code>author</code>
 									</td>
 									<td>The author of the publication.</td>
-									<td>One or more <a href="https://schema.org/Person"><code>Person</code></a> or <a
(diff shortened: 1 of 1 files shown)


iherman:
I hope this merge will work out with the inker entry in the other PR:-)

Let us merge these three PR-s as soon as possible to avoid conflicts...

PR #301, opened 2018-08-10
update links description to try and better clarify uses

mattgarrish:
Let me know if this makes things any better (or worse). I separated out the "linked to" use cases from the "resources not to be taken offline or packaged" to try and make it less confusing.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1930,30 +1930,32 @@
 				<section id="links">
 					<h4>Links</h4>
 
-					<p>The <dfn>links</dfn> property enumerates <a href="#wp-resources">resources</a> that are
-						significant to a <a>Web Publication</a> but are <em>not</em> essential to the processing and
-						rendering of it (i.e., the content of the Web Publication remains readable even if the resources
-						are not available). Examples of linked resources might include:</p>
+					<p>The <dfn>links</dfn> property provides a list of <a href="#wp-resources">resources</a> that are
+							<em>not</em> required for the processing and rendering of a Web Publication (i.e., the
+						content of the Web Publication remains unaffected even if these resources are not available).
+						Linked resources are typically made available to user agents to augment or enhance the
+						processing or rendering of it, such as:</p>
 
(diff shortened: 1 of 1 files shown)


llemeurfr:
This makes things really clearer. Only the last part is still fuzzy IMO:
 
> User agents MAY ignore linked resources, and are not required to take them offline with a Web Publication. These resources SHOULD NOT be included when packaging a Web Publication.

... means that the author cannot expect that links will survive offlining and packaging. This is consistent with the description of what links are. 

> The completeness of the links list might affect the usability of the Web Publication in certain scenarios (e.g., if a user agent does not take them offline). For this reason, it is strongly RECOMMENDED to provide a comprehensive list of all of the Web Publication's non-critical resources.

... could mean many obscure things. What is the connection between completeness of the list of links and offline reading? (if a link is not present, it's not present online and offline).  Why is this sentence useful? I would remove it.



... could mean that authors can expect that

mattgarrish:
> ... means that the author cannot expect that links will survive offlining and packaging.

I'm not seeing that. It refers to the resources being optional to take offline or package. We don't say to scrub the infoset when going offline, or not to include the manifest in the package, so there should always be persistence that way.

> Why is this sentence useful? I would remove it.

It was copied over from the resource list -- where it also makes more sense if you fail to list resource that are needed offline. At this point, given how we're stressing non-essential for links, it does seem a bit contradictory. I'm not sure the lack of a font will seriously affect usability. Were there other cases you had in mind @iherman, or do you think it can be dropped?

llemeurfr:
@mattgarrish my bad, I didn't mean that the link would disappear when offlining the content, but that the referenced content will not be fetched, therefore will not be accessible offline. Forget it.

iherman:
@mattgarrish yes, this seems to be copied from resources, and I agree it can be removed.

mattgarrish:
Any thoughts on this before merging, @dauwhe?

PR #304, opened 2018-08-13
Added infos on single values vs arrays and strings vs objects

### LINKED ISSUE #287: Feedback while working on a JSON Schema for WP
Since we won't have any calls for the next two weeks, I've started working on a JSON Schema for WP.

A first draft is available at https://github.com/w3c/wpub/blob/master/schema/publication.schema.json

While working on this schema, I've had to go through the specification again and it highlighted several issues with the current draft.

**1. Creators**

The JSON serialization for creators feels very much underdefined: there's a good list of roles that I can use for my JSON Schema but everything else feels underdefined.

- the required value is either "One or more Person" or "One or more Person or Organization"
- does this mean that we MUST use an object? My audiobook example used strings as well, since it's much easier to author it that way when you only have a name to indicate.
- do we require any terms in that object (`name`, `@type`) or have a strong preference for specific terms?
- do we allow arrays as well? Arrays of strings? Arrays of objects?

**2. Item-specific language**

I find that part of the spec every bit as confusing as creators:

- where exactly do we allow the use of `@value` + `@language` instead of a string? An explicit list would help a lot.
- having examples for every term where this is allowed would also help, ideally in each respective section rather than under its own section
- do we allow multiple localizations of each string as well (an array of objects)? 

**3. String/Array/Object**

Overall, it's fairly difficult to know right now where we allow strings, arrays or objects.

For example:

- do we allow all three for `name` (string for the simple case, object for a single localized value, array of objects for multiple localizations)?
- what about the different accessibility related terms? On the schema.org website, I see that some of them are strings, while others are arrays of strings. It would be useful to indicate this in our specification.

**4. Accessibility**

There are many different values for the various accessibility terms that we support, but they're not listed in our specification.
Should we validate these values at all? 

For languages (both `inLanguage` and `@language`), `readingProgression` and `inDirection`, I've already handled the validation of their values.

iherman:
The examples have also been extended for the various cases.

This answers to 
https://github.com/w3c/wpub/issues/287#issuecomment-412120871

Fixes #287

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/304.html" title="Last updated on Aug 13, 2018, 2:35 PM GMT (55e26e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/304/275d901...55e26e3.html" title="Last updated on Aug 13, 2018, 2:35 PM GMT (55e26e3)">Diff</a>

### DIFF
--- a/experiments/separate_manifest/mobydick-simple.jsonld
+++ b/experiments/separate_manifest/mobydick-simple.jsonld
@@ -32,7 +32,7 @@
         },{
             "@type": "PublicationLink",
             "url": "html/toc.html",
-            "rel": "contents",
+            "rel": "contents"
         },{
             "@type": "PublicationLink",
             "url": "fonts/STIXGeneral.otf",
--- a/experiments/w3c_rec/full_version.html
+++ b/experiments/w3c_rec/full_version.html
@@ -10,6 +10,7 @@
         "@type"                 : "CreativeWork",
         "@id"                   : "http://www.w3.org/TR/tabular-data-model/",
         "url"                   : "http://www.w3.org/TR/2015/REC-tabular-data-model-20151217/",
+        "accessibilityAPI"      : "ARIA",
         "accessMode"            : ["textual", "visual"],
         "accessModeSufficient"  : ["textual"],
         "copyrightYear"         : "2015",
@@ -82,29 +83,6 @@
                 "url"            : "test-utf8.csv",
                 "encodingFormat" : "text/csv"
 
-            },
-            {
-                "@type"          : "StructuredValue",
-                "url"            : "test-utf8-bom.csv",
-                "encodingFormat" : "text/csv"
-
-            },
-            {
-                "@type"          : "StructuredValue",
-                "url"            : "test-utf16.csv",
-                "encodingFormat" : "text/csv"
-
-            },
-            {
-                "@type"          : "StructuredValue",
-                "url"            : "test-utf16-bom.csv",
-                "encodingFormat" : "text/csv"
-
-            },
-            {
-                "@type"          : "StructuredValue",
-                "url"            : "test.xls",
-                "encodingFormat" : "application/vnd.ms-excel"
             },
             {
                 "@type"          : "StructuredValue",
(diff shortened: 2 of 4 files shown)


mattgarrish:
Content-wise I don't have any issue, but am wondering about organization. The new sections come before the section that references the properties, and link values are described in section 4. We're getting a bit all over the place.

I wonder if we should group them together, maybe like this:

```
2.3.3. Properties
2.3.3.1 Arrays and Single Values
2.3.3.2 Text Values and Objects
2.3.3.3 Links
2.3.3.3.1 PublicationLink Definition
```

?

iherman:
Yes. I was not sure either. The problem is, however, that the section on publication types (ie, 3.2.4) actually makes use of the fact that the same property value can be an array or a single value, hence 3.2.2. preceding it...)

Do you think there is a better place for the current 3.2.2 and 3.2.3?

(B.t.w., the section numbers in your comments are odd; we are talking about section 3, right?)

mattgarrish:
> the section numbers in your comments are odd; we are talking about section 3, right?

Yes, my ability to think is being disrupted by the sound of my neighbour sawing down trees.

> the section on publication types (ie, 3.2.4) actually makes use of the fact that the same property value can be an array or a single value

The sections currently only refer to properties in section 4. Maybe that needs generalizing, too, then?

Maybe we can make a "Values" section that precedes the type declaration, or:

```
3.2.2. Values
3.2.2.1 Arrays and Single Values
3.2.2.2 Text Values and Objects
3.2.2.3 Links
3.2.2.3.1 PublicationLink Definition
```

mattgarrish:
I merged the examples into a single unit with the last commit just to connect the thread a little more obviously. See if it all still makes sense.

iherman:
Matt, I like what you did. I have made a commit with 2 minor editorial glitches (all mine); I would propose to merge this; after all, this is a purely editorial change.

HadrienGardeur:
I'm still uncomfortable with allowing a single string in `readingOrder`, `resources` or `links`, which is why I'm not a big fan of these examples.

I'm fine with providing this kind of flexibility to metadata, but I don't think we should allow them for more structural properties.

PR #326, opened 2018-09-05
Add `doJsonLd: true` respec config opt

BigBlueHat:
See: https://github.com/w3c/respec/wiki/doJsonLd

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/326.html" title="Last updated on Sep 5, 2018, 7:01 PM GMT (c77fc73)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/326/12a9993...c77fc73.html" title="Last updated on Sep 5, 2018, 7:01 PM GMT (c77fc73)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,7 +46,8 @@
 				  repoURL: "https://github.com/w3c/wpub/",
 				  branch: "master"
 			  },
-              localBiblio: biblio
+              localBiblio: biblio,
+              doJsonLd: true
           };
           // ]]>
       </script>


iherman:
@BigBlueHat this sounds like a great idea (I must admit I did not know about this respec option), but I have one problem.

I have a script (that may have to be updated) that generates a Web Publication Manifest into the header, under the heading of "eat your own dog food". (Look at `common/js/wp.js`). However, I had to switch it off in the last publication round, because it did use the `rel=publication` link in the header (per our draft) and that rel was refused by our publication checker as not IANA defined. One possibility would be, rather than using the respec schema.org, to put this script back _without_ the additional link (and then have a separate discussion with the webmaster on getting a dispensation for this).

WDYT?

iherman:
@BigBlueHat, see https://github.com/w3c/echidna/issues/596 ; I think that will be solved eventually. Ie, the generated JSON-LD will be the Web Manifest.

Is it o.k. to close this PR?

TzviyaSiegman:
@BigBlueHat should this be merged or deleted?

mattgarrish:
It looks like the request got rejected, although no explanation was given, as the issue was closed and rel=publication is still reported as an error.

But we can still insert the manifest metadata without the link, which is closer to what we want, so let's close this off. Now that we know this flag exists, we can always implement it later if need be. Feel free to reopen if anyone disagrees.

Issue #334, opened 2018-09-12
primary entry page must be added to reading order when reading order not specified


mattgarrish:
[As discussed](https://github.com/w3c/wpub/pull/331#issuecomment-419635650) in PR #331, the default reading order is required, but the primary entry page may be omitted from it when it is the only resource it would contain. Missing is a statement that, in this case, user agents must add the primary entry page to the default reading order when compiling the infoset.

HadrienGardeur:
Isn't that covered in step 6 of [the canonical manifest algorithm](https://w3c.github.io/wpub/#canonical-manifest)?

mattgarrish:
Yes, but there should be a clearer statement in 4.4.1.1 that matches the expected behaviour. All it says right now is that the entry page can be omitted.

mattgarrish:
This is closed with #335.

Issue #333, opened 2018-09-12
manifest must be embedded in primary entry page


mattgarrish:
[As discussed](https://github.com/w3c/wpub/pull/331#issuecomment-419456205) in PR #331, there are currently no restrictions on what resource can contain the embedded manifest. Proposal is to restrict to the primary entry page (must provide a link to the manifest or must embed the manifest).

dauwhe:
Yes. Absolutely!

GarthConboy:
And semi-related... how does a WP-aware RS/AU know that a non-primary-entry-page resource is part of a WP.  E.g., a link to "chapter 27" -- how does such a RS/UA know to jump into "WP mode" following that link?  And, yea, all the "part of multiple WPs" issues arise too.

BigBlueHat:
If we want any SEO value from the contents of the manifest, it MUST be embedded in the entry page.

mattgarrish:
This is closed with #335.

Issue #329, opened 2018-09-06
The primary entry page may be under specified


llemeurfr:
From reading the specification, it is difficult to have an idea about what the primary entry page is made for. 

It's introduced in 3.5 resources as the only required resource in a WP. 
In 3.6 we discover that the ToC should be located there (if not the manifest should reference it). This is detailed in 4.6.2.
In 4.3.2 we discover that the Web Publication address is the URL of this resource (but 4.3.2.1 states that it may not be an HTML resource, in which case the user should not have access to it (SVG forbidden)).
In 4.3.5 we see that it can have some influence on the language and base direction of the WP, and in 4.3.9 that the title of the WP may be inferred from it. 

I wonder if such an important piece of information for WP should have its specification grouped in a specific section (in 3.5 ?).

iherman:
I see the point, I have actually hit this issue at some point; it is an important concept and its role is scattered over the document. Yes, it may be better to regroup things at one place.

@mattgarrish ?

mattgarrish:
I actually thought we had a section on it, but I probably just got confused by all the bikeshedding about the name we did. Certainly seems useful to detail it under the construction section.

llemeurfr:
@mattgarish yes let's close it, the new section is a really good start. Thanks.

PR #335, opened 2018-09-19
primary entry page issues

mattgarrish:
This PR addresses the following three issues on the primary entry page (pep):
 
- #329 - that we should have a dedicated section to explain the pep
- #333 - that the manifest must be embedded in the pep
- #334 - that the pep has to be added to the default reading order when the reading order isn't specified

Feedback on whether I've captured the intent of the page correctly welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/335.html" title="Last updated on Sep 19, 2018, 12:08 PM GMT (f7056b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/335/8521f06...f7056b7.html" title="Last updated on Sep 19, 2018, 12:08 PM GMT (f7056b7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -556,11 +556,10 @@
 			<section id="manifest-embedding">
 				<h3>Embedding a Manifest</h3>
 
-				<p>A manifest can be embedded within an HTML document using the <a
+				<p>When opting to embed the manifest, it MUST be included in the <a href="#wp-primary-entry-page"
+						>primary entry page</a> using the <a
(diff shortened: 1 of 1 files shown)


iherman:
There may be some details to improve, but it would be important to merge this PR and take it from there. It covers a major missing editorial aspect of the document which may be the source of misunderstandings.

Issue #338, opened 2018-10-08
Simplifying the WebIDL


HadrienGardeur:
The current WebIDL has dedicated elements for a few infoset items:

- [accessibility report](https://github.com/w3c/wpub/blob/master/webidl/manifest.webidl#L43)
- [privacy policy](https://github.com/w3c/wpub/blob/master/webidl/manifest.webidl#L44)
- [cover](https://github.com/w3c/wpub/blob/master/webidl/manifest.webidl#L45)
- [table of contents](https://github.com/w3c/wpub/blob/master/webidl/manifest.webidl#L46)

I think that having these in our WebIDL is a little bit misleading:

- they mostly duplicate info that are in `readingOrder`, `resources` or `links`
- I don't think we have any specific handling in them in our [canonicalization](https://w3c.github.io/wpub/#canonical-manifest) or [processing of the manifest](https://w3c.github.io/wpub/#processing-manifest), which means that currently they would all be empty
- they're all helpers, which are usually meant to facilitate fetching specific info in a model, which would be fine as methods in an object, but IMO are not meant to show up in IDL

iherman:
@HadrienGardeur I do not have a strong opinion on this. I understand the arguments. For me the central point is your third one:

> they're all helpers, which are usually meant to facilitate fetching specific info in a model, which would be fine as methods in an object, but IMO are not meant to show up in IDL

yes indeed, they are helpers, and they would have their place as a method in a, say, JS object that embodies the Manifest. We agree on that. However, I tend to view our WebIDL (because we do not define some sort of a Web API) as mostly an abstract, language independent definition of such an object and, hence, I think it is better to keep it there. But, as I say, I do not have a strong opinion, ie, I am fine with whatever direction the WG takes.

@TzviyaSiegman @wareid @GarthConboy 

(Note, for completeness, that if the PR #339 is merged, then `pagelist` should be added to the list.)

HadrienGardeur:
@iherman this is my first time working with WebIDL, so I'm clearly no expert at this but from a quick glance at the spec, if we really want to declare such helpers, they should be defined as getters, not the way they're currently defined where we're basically duplicating info.

Do we have anyone with good knowledge of the way WebIDL works in our group?

iherman:
[...]

PR #339, opened 2018-10-09
Added the pagelist entry

### LINKED ISSUE #223: Other navigation elements beyond ToC

iherman:
This is the action recorded in [minutes of 8th of October](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2018/2018-10-08-pwg.html#resolution2).

Some notes

- @avneesh: I have added a text in 3.8 that I copied, essentially, from the similar entry in the EPUB document. I wonder if the pagelist as we have here would not be a bit more general, and not depend on "page boundaries present in a print source". A more general text would be therefore welcome.
- @HadrienGardeur: I am aware of #338, but I decided to keep this PR independent of that one, ie, follow the current draft and add the pagelist entry to the WebIDL. If that issue leads to overall changes we can remove that easily, but my goal was to keep things consistent.

If this PR is merged, then https://github.com/w3c/wp-vocab/pull/1 should also be merged and processed; this adds the extra `rel` value to our vocabulary files.

Cc: @GarthConboy @llemeurfr @laudrain @HadrienGardeur 

Fix #223

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/339.html" title="Last updated on Oct 12, 2018, 1:14 PM GMT (a176c0a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/339/2dfc275...a176c0a.html" title="Last updated on Oct 12, 2018, 1:14 PM GMT (a176c0a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -85,8 +85,8 @@
 					into a single contiguous work.</p>
 
 				<p>A Web Publication is discoverable in one of two ways: resources either include a link to the manifest
-					(via an HTTP Link header or an HTML <code>link</code> element&#160;[[html]]), or the manifest can
-					be loaded directly by a compatible user agent.</p>
+					(via an HTTP Link header or an HTML <code>link</code> element&#160;[[html]]), or the manifest can be
+					loaded directly by a compatible user agent.</p>
 
 				<p>With the establishment of Web Publications, user agents can build new experiences tailored
 					specifically for their unique reading needs.</p>
@@ -680,9 +680,12 @@
 				<p>The table of contents provides a hierarchical list of links that reflects the structural outline of
 					the major sections of the Web Publication.</p>
 
-				<p>
-					The table of contents is expressed via an HTML element (typically a <code>nav</code> element&#160;[[!html]]) in one of the <a href="#wp-resources">resources</a>. This element MUST be identified by the <code>role</code> attribute&#160;[[!html]] value "<code>doc-toc</code>"&#160;[[!dpub-aria-1.0]], and MUST be the first element in the document with that <code>role</code> value <a href="https://dom.spec.whatwg.org/#concept-tree-order">document tree order</a>&#160;[[!dom]].
-				</p>
+				<p> The table of contents is expressed via an HTML element (typically a <code>nav</code>
+					element&#160;[[!html]]) in one of the <a href="#wp-resources">resources</a>. This element MUST be
+					identified by the <code>role</code> attribute&#160;[[!html]] value
+					"<code>doc-toc</code>"&#160;[[!dpub-aria-1.0]], and MUST be the first element in the document with
+					that <code>role</code> value in <a href="https://dom.spec.whatwg.org/#concept-tree-order">document
+						tree order</a>&#160;[[!dom]]. </p>
 
 				<p>If the table of contents is not located in the <a href="#wp-primary-entry-page">primary entry
 						page</a>, the manifest SHOULD <a href="#table-of-contents-manifest">identify the resource</a>
@@ -702,25 +705,27 @@
 			<section id="wp-pagelist">
 				<h3>Pagelist</h3>
 
-				<p>
-					The pagelist is a list of links that provides navigation to static page demarcation points within the content. These locations allow users, for example, to coordinate access into the content. The exact nature of these locations is left to content creators to define. They usually correspond to pages of a print document which is the source of the digital publication, but might be a purely digital creation added for the sake of easing navigation.
-				</p>
+				<p> The pagelist is a list of links that provides navigation to static page demarcation points within
+					the content. These locations allow users, for example, to coordinate access into the content. The
+					exact nature of these locations is left to content creators to define. They usually correspond to
+					pages of a print document which is the source of the digital publication, but might be a purely
+					digital creation added for the sake of easing navigation. </p>
 
-				<p>
-					The pagelist is expressed via an HTML element (typically a <code>nav</code> element&#160;[[!html]]) in one of the <a href="#wp-resources">resources</a>. This element MUST be identified by the <code>role</code> attribute&#160;[[!html]] value "<code>doc-pagelist</code>"&#160;[[!dpub-aria-1.0]], and MUST be the first element in the document with that <code>role</code> value <a href="https://dom.spec.whatwg.org/#concept-tree-order">document tree order</a>&#160;[[!dom]].
-				</p>
+				<p> The pagelist is expressed via an HTML element (typically a <code>nav</code> element&#160;[[!html]])
+					in one of the <a href="#wp-resources">resources</a>. This element MUST be identified by the
+						<code>role</code> attribute&#160;[[!html]] value
+					"<code>doc-pagelist</code>"&#160;[[!dpub-aria-1.0]], and MUST be the first element in the document
+					with that <code>role</code> value <a href="https://dom.spec.whatwg.org/#concept-tree-order">document
+						tree order</a>&#160;[[!dom]]. </p>
 
-				<p>
-					If the pagelist is not located in the <a href="#wp-primary-entry-page">primary entry page</a>, the manifest SHOULD <a href="#pagelist-manifest">identify the resource</a> that contains the structure.
-				</p>
+				<p> If the pagelist is not located in the <a href="#wp-primary-entry-page">primary entry page</a>, the
+					manifest SHOULD <a href="#pagelist-manifest">identify the resource</a> that contains the structure. </p>
 
-				<p>
-					There are no requirements on the pagelist itself, except that, when specified, it MUST include a link to at least one <a href="#wp-resources">resource</a>.
-				</p>
+				<p> There are no requirements on the pagelist itself, except that, when specified, it MUST include a
+					link to at least one <a href="#wp-resources">resource</a>. </p>
 
-				<p>
-					Refer to the <a href="#pagelist">pagelist property definition</a> for more information on how to identify in the infoset and manifest which resource contains the pagelist.
-				</p>
+				<p> Refer to the <a href="#pagelist">pagelist property definition</a> for more information on how to
+					identify in the infoset and manifest which resource contains the pagelist. </p>
 			</section>
 		</section>
 		<section id="wp-properties">
@@ -760,8 +765,8 @@
 					<dt><a href="#structural-properties">structural properties</a></dt>
 					<dd>
 						<p>Structural properties identify key meta structures of the Web Publication, such as the <a
-								href="#cover">cover image</a>, or the location of the <a href="#table-of-contents"
-								>table of contents</a> or the <a href="#pagelist">pagelist</a>.</p>
+								href="#cover">cover image</a>, or the location of the <a href="#table-of-contents">table
+								of contents</a> or the <a href="#pagelist">pagelist</a>.</p>
 					</dd>
 				</dl>
 
@@ -912,8 +917,8 @@
 							<td><a href="#cover">Cover</a></td>
 						</tr>
 						<tr>
-								<td><code>https://www.w3.org/ns/wp#pagelist</code></td>
-								<td><a href="#pagelist">Pagelist</a></td>
+							<td><code>https://www.w3.org/ns/wp#pagelist</code></td>
+							<td><a href="#pagelist">Pagelist</a></td>
 						</tr>
 						<tr>
 							<td><code>id</code></td>
@@ -2570,17 +2575,24 @@
 
 						<ol>
 							<li>Identify the table of content resource: <ul>
-									<li>
-										If a resource in either the <a href="#default-reading-order">default reading order</a> or <a href="#resource-list">resource-list</a> is identified with a <code>rel</code> value including <code>contents</code>&#160;[[!iana-link-relations]], the corresponding <code>url</code> value identifies the table of content resource. If there are several such resources, the first one MUST be used, with the <a href="#default-reading-order">default reading order</a> taking precedence over <a href="#resource-list">resource-list</a>.
-									</li>
-									<li>
-										Otherwise, the <a>primary entry page</a> is the table of content resource.
+									<li> If a resource in either the <a href="#default-reading-order">default reading
+											order</a> or <a href="#resource-list">resource-list</a> is identified with a
+											<code>rel</code> value including
+										<code>contents</code>&#160;[[!iana-link-relations]], the corresponding
+											<code>url</code> value identifies the table of content resource. If there
+										are several such resources, the first one MUST be used, with the <a
+											href="#default-reading-order">default reading order</a> taking precedence
+										over <a href="#resource-list">resource-list</a>. </li>
+									<li> Otherwise, the <a>primary entry page</a> is the table of content resource.
 									</li>
 								</ul>
 							</li>
-							<li>
-								If the table of content resource contains an HTML element with the <code>role</code>&#160;[[!html]] value <code>doc-toc</code>&#160;[[!dpub-aria-1.0]], the user agent MUST use that element as the table of contents. If there are several such HTML elements the user agent MUST use the first in <a href="https://dom.spec.whatwg.org/#concept-tree-order">document tree order</a>&#160;[[!dom]].
-						</li>
+							<li> If the table of content resource contains an HTML element with the
+								<code>role</code>&#160;[[!html]] value <code>doc-toc</code>&#160;[[!dpub-aria-1.0]], the
+								user agent MUST use that element as the table of contents. If there are several such
+								HTML elements the user agent MUST use the first in <a
+									href="https://dom.spec.whatwg.org/#concept-tree-order">document tree
+								order</a>&#160;[[!dom]]. </li>
 						</ol>
 
 						<p>If this process does not result in a link to the table of contents, the Web Publication does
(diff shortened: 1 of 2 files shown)


HadrienGardeur:
@iherman aside from my comment in #338, I don't think that `pagelist` and `toc` can simply return an `HTMLElement`.

Since they can both be contained in resources other than the entry page, we need to know the URIs of such resources as well.
There's also the case of multiple resources marked as being the TOC and/or page liste as well, which requires extra attention.

HadrienGardeur:
Sidenote: we can usually preview such PRs but I'm not seeing any link for that right now. Is something broken in our pipeline?

iherman:
I left this intentionally open for now, because the whole issue of ToC is still open as far as I am concerned, and until we have a clear idea of what we want, I do not want to touch the WebIDL. That should come when all other issues around ToC are closed.

For pagelist, I think we should, for now, simply keep it as a clone of ToC, so to say, and handle them together. This should _not_ be done as part of this PR imho...

HadrienGardeur:
By the way, all of the previous comment is also true for the TOC and this is completely independent from any discussion about the machine readability of such documents.

avneeshsingh:
How would this approach work in a non-WP aware user agent. It would be good to know the details.

lrosenthol:
@avneeshsingh great question!  If we use data attributes, then the non-WP-aware UA would ignore them (since that's the rules - https://www.w3.org/TR/html52/dom.html#embedding-custom-non-visible-data-with-the-data-attributes).  However, because they are "connected" with the content (being attributes), they *could* be used by any HTML UA, such as an assistive technology device, even if it wasn't WP aware.

danielweck:
@lrosenthol isn't that abusing / misusing custom `data-xxx` attributes?

"User agents must not derive any implementation behavior from these attributes or values. Specifications intended for user agents must not define these attributes to have any meaningful values."

"Custom data attributes are intended to store custom data private to the page or application, for which there are no more appropriate attributes or elements."

"These attributes are not intended for use by software that is not known to the administrators of the site that uses the attributes. For generic extensions that are to be used by multiple independent tools, either this specification should be extended to provide the feature explicitly, or a technology like microdata should be used (with a standardized vocabulary)."

"JavaScript libraries may use the custom data attributes, as they are considered to be part of the page on which they are used."

etc.
https://www.w3.org/TR/html52/dom.html#embedding-custom-non-visible-data-with-the-data-attributes

danielweck:
> The pagelist is a list of links that provide navigation to positions in the content that correspond to the locations of page boundaries present in a print source being represented by the Web Publication

I am wondering whether it would be useful to clarify "print source" here, as this could be interpreted exclusively as "publications printed on paper, like traditional books or magazines", when this could in fact be referring to digital publications *not* designed to be fed through a printer at all (i.e. born-digital, stay-digital documents), such as a fixed-layout / pre-paginated  EPUBs (or perhaps even PDFs) with interactive features, videos, animations, etc. I have seen a few examples of sales brochures, technical documentation, children's talking books, cooking / recipe books, etc. that do not have print equivalent, yet they are "paginated" works that would benefit from "pagelist" navigation in Web Publications.

lrosenthol:
@danielweck 
> isn't that abusing / misusing custom data-xxx attributes?

Probably...but HTML's limited model for attributes doesn't give us too much other choice..short of going to the HTML committee and asking for a new one.  But I still think either of those is a better choice than creating a completely new grammar/language that requires custom parsing...

danielweck:
> Pagelist is important for Web Publications with print equivalent, however it may also be used for Web Publications which do not have print equivalent, for making the navigation more convenient.

I am not sure I understand the intent of this additional prose (which was inserted on top of the original proposal), and I am certainly not convinced by the claim "for making the navigation more convenient" which does not seem based upon a well-defined rationale.

Normally, we start with use cases / functional requirements, and if we agree on the needs, we translate that to well-specified / standardized mechanisms. With the additional prose I feel that we are introducing a tautological statement, and an imprecise one as well. Here is how I read the current prose: "page lists are important for publications that somehow have pages, but even if they don't, this feature is useful too because it provides navigation convenience").

Let's look at it from a different angle, and let's say that we're now debating the inclusion of EPUB's "list of landmarks" (e.g. figures, mathematical equations, etc.) in the Web Publications standard. Just like "pagelist", there would be a well-defined metadata vocabulary (or semantic role) that reading systems / user agents would use to discover the list of links identifying worthy locations within publication resources. It would be strange to state that "this feature can also be used for publications that do not actually contain figures, to make navigation more convenient".

I am sorry if I am missing the point :)
Perhaps I am just nit-picking.

iherman:
@danielweck : as I said on the call yesterday, I do need a text for the paragraph referring you, and I welcome any input:-)

iherman:
@danielweck maybe the discussion should move back to #223...

danielweck:
> @danielweck maybe the discussion should move back to #223...

You mean the discussion about `data-xxx` attributes, right? (+ @lrosenthol )

I think this PR is the right place to discuss the prose proposed in this PR :)

TzviyaSiegman:
@lrosenthol  granular anchoring is not the topic of discussion here. Please open a separate issue if you'd like to see that addressed.

iherman:
@danielweck no. #223 is on any other navigation elements, including the pagelist. This PR implements the current resolution of including pagelist, but if we question this resolution, or want to extend it to other categories, the discussion should be done there, imho.

lrosenthol:
@TzviyaSiegman I disagree, as we are discussing the pagelist here...However, it appears that @iherman thinks that #223 would be a better spot and I am happy to move it there since both @danielweck and I have concerns about the currently proposed pagelist mechanism .

danielweck:
> as I said on the call yesterday, I do need a text for the paragraph referring you, and I welcome any input:-)

@iherman 

(1) I am proposing a clarification of "print source" as per my comment here ( https://github.com/w3c/wpub/pull/339#issuecomment-428262452 ) which I feel is neither an objection to the resolution, nor a request for extending the semantics of "page list" (perhaps the latter is where our interpretation differ, in which case I will be happy to move to #223 ).

(2) Regarding the additional sentence "Pagelist is important for Web Publications with print equivalent, however it may also be used for Web Publications which do not have print equivalent, for making the navigation more convenient", right now I am leaning towards removing it entirely, based on the reasons mentioned in comment https://github.com/w3c/wpub/pull/339#issuecomment-428269828 (I feel that the original prose clearly addresses the identified use-case / functional requirement, especially if the notion of "print source" is clarified)

mattgarrish:
I agree with @danielweck that the wording right now needs improvement. What about something like:

> The pagelist is a list of links that provides navigation to static page demarcation points within the content. These locations allow users, for example, to coordinate access into the content. The exact nature of these locations is left to content creators to define. They might correspond to a statically paginated source, for example, such as a print document, or might be a purely digital creation to simplify access into the content.

How we provide the convenience of pagination in a digital world is a problem that needs a better solution some day, but abusing attributes that we know are just going to lead to non-implementation is not the way forward. It's not a problem we're likely to solve on WP's timeline.

TzviyaSiegman:
I agree with @danielweck 's concerns
1. "print source" is a bit unclear. Perhaps we should phrase it as "paginated content" or simply remove the phrase. HTML can be paginated too. 
2. +1 to removing mention of when it might be helpful to use pagelist. We leave this as a MAY and allow the user to determine when to use it.

danielweck:
To put it differently: once the notion of "print source" is clarified to encompass any type of publication that intrinsically provides / contains a "list of pages" (thus the "pagelist" keyword / semantic role proposed here), such as "printed physical publications", "digital fixed-layout publications", and otherwise "reflowable publication with some sort of virtual pagination markers" (which I believe was the primary motivator for the additional prose), then we can remove the additional prose.

danielweck:
@mattgarrish I like the term "static" because I interpret its opposite meaning ("dynamic") as "pagination at rendering time, not baked into the content at authoring time". I am not sure everybody would use the same interpretation though.

iherman:
I like the formulation in https://github.com/w3c/wpub/pull/339#issuecomment-428289554. I see only one caveat, though: this formulation makes it feasible to define _several_ panelist-s, doesn’t it? I do not think this is a problem per se, but, if this is indeed the case:

* are we sure the term “panelist” is the right one? Shouldn’t we use a more general term? (Maybe not if we stick to the DPUB ARIA label)
* the current text relies on the approach related to the ToC that talks explicitly of the case when several pointers to lists are added to the manifest. We may want to remove that
* we may suggest the usage of the `PublicationLink` structure adding the `name` field categorizing/describing what that particular panelist is used for (do we need some predefined terms to be put there?)

mattgarrish:
> we may suggest the usage of the `PublicationLink` structure adding the `name` field categorizing/describing what that particular panelist is used for

Technically, there's no reason why you couldn't have more than one page list, but realistically coordinating the locations is usually only accomplished against one source (or using one "distance" method if there isn't a paginated source/equivalent).

I'd be fine with allowing more than one and using the `name` field as you suggest, but I don't have a strong opinion on it.

With regards to the name, I can't think of a better alternative. We have page lists in EPUB and DPUB-ARIA, so there's consistency in comprehension if we don't alter the name. The concept of getting to a "page", even if the user doesn't fully understand the pagination method, is also broadly understood.

llemeurfr:
The first part of Matt's proposal is clear:

> The pagelist is a list of links that provides navigation to static page demarcation points within the content. These locations allow users, for example, to coordinate access into the content.

The second part, using "might", is too generic, its blurs the notion: it's a `page list` after all, not a random list of pointers. I would rephrase/simplify it as: 

> The exact nature of these locations is left to content creators to define. They usually correspond to the pages of a print document which is the source of the digital publication, but might be a purely digital creation adding to the ease of navigation.

Plus, what is missing is the exact meaning of the page break in the content, vs start/end of the printed page. The [Daisy page on page lists](http://kb.daisy.org/publishing/docs/navigation/pagelist.html ) adds it in a Q&A section (=start). 

Last, the representation of the page number in the content should IMO be specified also: EPUB is underspecified, which is an issue for UAs, and the Daisy page also gives details on best practices.

avneeshsingh:
@llemeurfr

I think we should focus on mechanism in specs here, and leave the details of start or end of page to best practices. there are many types of publications which would be having different implementations of page numbers. Some may be numeric, some may be alphabetic and some may have labels. Some print equivalent may have page numbers marked at top of page, some may have it at bottom.
DAISY 2 specs were precise about it, so production centers had to implement work arounds to match print equivalent.

mattgarrish:
> Last, the representation of the page number in the content should IMO be specified also

Ideally, yes, but practically this can't be enforced without unwanted side-effects. You can have a page list for an audio book, for example, which would only reference into audio offsets. The page list itself is only dependent on having some location to link to.

llemeurfr:
In the proposed text I read [the pagelist nav]  "MUST be the first element in the document so designated." It seems impossible to have both the ToC nav and the pagelige nav in the same entry page, both being the first element in the document. Or Do I miss something?

mattgarrish:
> I'll create a specific issue for the issue of page numbers as content or empty elements.

Oh, okay, I misunderstood your request. One common issue is whether the numbers have to be present with the page list and the other is whether they have to be content or not. I can comment on the latter when you open the issue. :)

mattgarrish:
> It seems impossible to have both the ToC nav and the pagelige nav in the same entry page, both being the first element in the document.

The wording is a bit confusing, as it seems to mix authoring and user agent requirements. I read it as the pagelist/toc MUST be identified by the appropriate role, and user agents MUST recognize the first one so designated. It would probably be better to split these statements, if so.

iherman:
> In the proposed text I read [the pagelist nav] "MUST be the first element in the document so designated." It seems impossible to have both the ToC nav and the pagelige nav in the same entry page, both being the first element in the document. Or Do I miss something?

"so designated" includes the choice of the `nav` based on the value of `role` (`doc-toc` vs. `doc-panelist`). Ie, the choice of these two are disjoint, and they can both be in the same file.

iherman:
@HadrienGardeur 

> In the current PR, there's a contradiction between:
>
> * the spec language that allows the use of a rel to identify a page list in resources or readingOrder
> * the WebIDL which would only work if the page list is contained in the entry page

I do not see why the second statement would be true. It is an HTMLElement, ie, a DOM object, and that can be in any HTML resource, whether it is the entry page or not. 

But, again, this is the same as the ToC. Finalizing the ToC would fix this, too.

HadrienGardeur:
@iherman this means that in addition to having the DOM of the current document being displayed (for example chapter 1 instead of the entry page), you would need the DOM for potentially three additional documents as well: 

- the one containing the manifest
- the one containing the TOC
- the one containing the pagelist

This is quite heavy in terms of processing and I really don't think that's a good idea.

iherman:
Trying to find a consensus _on this PR_, based on the texts above. I extract the following possible changes before merge:

1. The text in [3.8](https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/339.html#wp-pagelist) should become:

> The pagelist is a list of links that provides navigation to static page demarcation points within the content. These locations allow users, for example, to coordinate access into the content. The exact nature of these locations is left to content creators to define. They usually correspond to pages of a print document which is the source of the digital publication, but might be a purely digital creation added for the sake of easing navigation.

2. in [3.7](https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/339.html#pagelist-infoset) and [3.8](https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/339.html#wp-pagelist) part of the sentence

> , and MUST be the first element in the document so designated.

should say

> , and MUST be the first element in the document with that `role` value.

---

I have also looked at the issue on whether we would allow several pagelists. This requires some "stylistic" changes ("A pagelist" instead of "The pageslist") and (2) above is moot for pagelists. However, it also raises the question on how to locate them. We could say:

1. The UA tries to locate pagelists in the primary entry page
2. _Additionally_, it will follow all entries, if available, in `resources` or `readingOrder`.

Do we agree to have this? Is it important to have several pageslists, in fact? Or should we forget about this for now?

---

Additionally, some other questions arose, to be discussed elsewhere, but should not prevent merging this PR:

- should there be definition or advise to identify page number content
- possible changes as a result of #338 would affect the WebIDL referring to pagelist
- have some text in the draft (or best practices) on what should kind of targets should be in pageslists
- usage of `data-xxx` attributes to identify targets

HadrienGardeur:
> Additionally, it will follow all entries, if available, in `resources` or `readingOrder`.

Yikes. Why would we do that? The pagelist should be identified in `resources` or `readingOrder` using a `rel` value.

iherman:
@HadrienGardeur, on https://github.com/w3c/wpub/pull/339#issuecomment-428872262 ("you would need the DOM for potentially three additional documents as well"): that is unfortunately correct. But this is related to the (still open) issue #291. If that was closed, we could simplify the WebIDL in some way, possibly defining a helping function that returns a ToC/pageslist in some JSON format for further processing. But, until that issue is open, we are stuck.

iherman:
@HadrienGardeur 

> Yikes. Why would we do that? The pagelist should be identified in resources or readingOrder using a rel value.

That is what I meant. If we have several pagelists, then each of them should be treated as we have defined for one.

Note that I do not have strong feelings whether we need this or not, ie, whether we should allow several pages lists or not. I was just exploring what this would mean.

avneeshsingh:
@iherm
The changes written by you are fine. Regarding multiple pagelist, I think it would be good to keep away from geting into to much details. We heard about a use case from Dave but we should know about more use cases to specify multiple pagelist properly.

HadrienGardeur:
> there are many types of publications which would be having different implementations of page numbers. Some may be numeric, some may be alphabetic and some may have labels.

@avneeshsingh IMO this is exactly what makes it very difficult for UAs to implement good features based on page lists.

For example, as a UA developer I can't show a dialog that will let the user jump to page "9" because:

- the pagelist might skip "9" entirely
- or it is indicated as "IX" or "ix"

As a UA developer, I can only provide:

- a long list of references (not super readable)
- and somehow offer the ability to display the page number for what I'm currently viewing (as an overlay)

I frankly don't think that's enough for education where we need to easily share a location AND jump to it as well (teacher tells the class to read page 9).

avneeshsingh:
In WP, we need to walk between flexibility and precision. This is why the best practices are so important. I would love to have it well defined that can address main use cases, but then it will restrict flexibility.

laudrain:
> and somehow offer the ability to display the page number for what I'm currently viewing (as an overlay)

In the content, page breaks have to identified with semantic, for instance the role "doc-pagebreak".

iherman:
Unless there is outcry, I intend to modify the PR today along the lines of https://github.com/w3c/wpub/pull/339#issuecomment-428881172, _except_ that I would keep the unicity of pagelists (ie, there should be only zero or one), following the remarks of [Avneesh](
https://github.com/w3c/wpub/pull/339#issuecomment-428886323).

avneeshsingh:
@iherman 
Good for now. If the group members find that multiple page lists are important, it would be good to open a new issue for it. I anticipate that identifying multiple pagelists with different print sources (as mentioned by Dave) will need additional work, so it would be good to document the use cases for it before we expand pagelist to pagelists.

index.html
```diff
@@ -2636,385 +2663,468 @@ <h5>Manifest Expression</h5>
...
+									Otherwise, the <a>primary entry page</a> is the pagelist resource.
+								</li>
+							</ul>
+						</li>
+						<li>
+							If the pagelist resource contains an HTML element with the <code>role</code>&#160;[[!html]] value <code>doc-pagelist</code>&#160;[[!dpub-aria-1.0]], the user agent MUST use that element as the pagelist.
+						</li>
+					</ol>
 
-				<section id="extensibility-linked-records">
```
mattgarrish: fix id

index.html
```diff
@@ -2636,385 +2663,468 @@ <h5>Manifest Expression</h5>
...
-					<h5>Linked records</h5>
+					<p>
+						If identifying the pagelist is ambiguous (e.g., several pagelist resources
```
mattgarrish: This paragraph contradicts the statement above that the the pagelist/toc must be the first one with the role. Here we don't say the user agent must use the first one, but rather that it can do what it chooses. I'd rather strengthen the statement here with respect to multiple identified elements and drop the must above, as it's not specifically a normative authoring statement. It might be better above to warn authors that only the first one is recognized.

index.html
iherman: I do not find the problem... what are you referring to?

index.html
```diff
@@ -2636,385 +2663,468 @@ <h5>Manifest Expression</h5>
...
+					<h5>Infoset Requirements</h5>
 
-				<p>Although both methods are valid, the use of linked records to extend the infoset is RECOMMENDED.</p>
+					<p>User agents MUST compute the <code>pagelist</code> as follows:</p>
 
-				<p>This specification does not define how such additional properties are compiled, stored or exposed by
-					user agents in their internal representation of the infoset. A user agent MAY ignore some or all
-					extended properties.</p>
+					<ol>
+						<li>Identify the table of content resource: <ul>
```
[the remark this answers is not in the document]
iherman: Done (in my machine for now:-)

index.html
iherman: My preference would be a bit the opposite... By actually remove this whole text (both here and in the ToC where it comes from). Yes, we do say that the UA must use the first one, and that is it...

iherman:
@mattgarrish I had to make some more changes to make the first-come-first-served rule:-) Please check. Maybe merge after this?

mattgarrish:
Looks good to me now.

PR #343, opened 2018-10-15
Reference to opaque origin in the manifest lifecycle

### LINKED ISSUE #321: What is an "opaque origin" and why do we care?
This comes up in obtaining a manifest and in the life cycle diagrams. But a trip to the HTML spec is no help, defining "opaque origin" as:

> An internal value, with no serialisation, for which the only meaningful operation is testing for equality.

How would a document end up having an opaque origin?

iherman:
Adopted the [change proposal](https://github.com/w3c/wpub/issues/321#issuecomment-429741645) of @JayPanoz. This replaces the (opaque:-) reference to an opaque origin...

Some issues on security must be addressed in the security section, too, which may, eventually, add some more text to the lifecycle section. But this PR seems to provide a proper intermediate stage.

Fix #321 

Cc: @JayPanoz, @rdeltour @llemeurfr @dauwhe

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/343.html" title="Last updated on Oct 15, 2018, 9:19 AM GMT (1349ceb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/343/1481110...1349ceb.html" title="Last updated on Oct 15, 2018, 9:19 AM GMT (1349ceb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2983,7 +2983,7 @@
 						obtaining a manifest</dfn>, starting from the <a>primary entry page</a>, are given by the
 					following algorithm. The algorithm, if successful, returns a <a>processed manifest</a>; otherwise,
 					it terminates prematurely and returns nothing. In the case of nothing being returned, the user agent
-					MUST ignore the manifest declaration. </p>
+					MUST ignore the manifest declaration.</p>
 
 				<ol>
 					<li> From the <code>Document</code> of the top-level browsing context of the <a>primary entry
@@ -2991,7 +2991,7 @@
 							link</var> be the first <code>link</code> element in tree order in <code>Document</code>
 						whose <code>rel</code> attribute contains the <code>publication</code> token. </li>
 
-					<li> If <var>origin</var> is an [[!html]] opaque origin, terminate this algorithm. </li>
+					<li> If the <code>Document</code> is sandboxed (e.g. a cross-origin document in an iframe&nbsp;[[!html]], terminate the algorithm.</li>
 
 					<li> If <var>manifest link</var> is <code>null</code>, terminate this algorithm. </li>
 


mattgarrish:
I'm not 100% sold on this, to be honest. Why select only one type of opaque origin to reject? data: URLs, srcdoc, createDocument() all seem like possible ways of generating an opaque-origin document that could attempt to initiate itself as a web publication, as we shouldn't only focus on what is common.

As I read the definition -- and I'm not claiming to be any kind of expert on this -- if a document has an opaque origin, its origin (```document.origin```) is set to NULL. Isn't that sufficient in terms of determining whether to terminate manifest processing? Only the browser developers really need to understand the finer details of calculating that origin.

JayPanoz:
Yeah that was my preferred option as well. 

That’s for implementers and It basically takes care of all the details for you, although the concept of “opaque origin” is complex to define/understand at first sight. My assumptions is that it will be re-introduced at a later date anyway, because it does that. 

Now, I can also understand the need for a temporary clarification, although I’m not comfortable with it – after all, it wasn’t at all in the W3C manifest at first and it could for instance buy some time to write some “companion doc” explaining how you may end up with an opaque origin and what it means in practice.

rdeltour:
I agree with @mattgarrish.

Also, I'm wondering what we're trying to solve with this PR?

- if the issue is that "opaque" isn't well-defined, then we should open an issue to the HTML standard instead.
- if the issue is that "opaque" is a concept difficult to comprehend, then I would think that replacing it with "sandboxed" wouldn't make it easier to understand for people not versed in the Web security model. In any case, as Matt correctly points out, this part of the spec is not for laypeople, but for UA developers.

The Web App Manifest issues referenced by @jaypanoz in #321 are on point (especially w3c/manifest#668). I think we should carefully review what happens there, and possibly seek advice from experts (at TPAC?), before jumping into changing anything here.

My 2 cents ;-)

JayPanoz:
@rdeltour Note opaque origin also impacts media elements as well, `audio` in particular. But I can’t tell what’s happening there if the media element is CORS cross-origin.

JayPanoz:
> but I would also think that the comment of @mattgarrish (ie, #343 (comment)) also holds, as an answer, to #321, which may have to be closed without further action, too

LGTM. 

(I did really think a lot about it recently but my view hasn’t change that much: I personally consider it a whole i.e. the Web Origin + its security policies so it’s probably more of an overarching concept.)

iherman:
Closing the PR without further action.

PR #353, opened 2018-10-24
Reading order and resources together set the bounds of a publication

### LINKED ISSUE #205: We need a section of the document that explicitly defines the bounds of a publication
We talk about the bounds of the publication but we never explicitly define what it means, where it comes from and what a UA is to do with it (and in what specific use cases).

iherman:
Made changes as [decided upon at the F2F](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2018/2018-10-22-pwg.html#section4) with the discussion also recorded in #205.

Fix #205.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/353.html" title="Last updated on Oct 24, 2018, 5:14 PM GMT (8d5c9b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/353/611cd98...8d5c9b7.html" title="Last updated on Oct 24, 2018, 5:14 PM GMT (8d5c9b7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -255,9 +255,9 @@
 
 					<p>To ensure interoperability when exposing the infoset items, this specification defines an
 						abstract representation of the data structures using the Web Interface Definition Language
-						(WebIDL) [[webidl-1]] which expresses the expected name, datatype, and possible restrictions
-						for each member of the manifest expressed in the infoset. (A WebIDL representation
-						can be mapped onto ECMAScript, C, or other programming languages.)</p>
+						(WebIDL) [[webidl-1]] which expresses the expected name, datatype, and possible restrictions for
+						each member of the manifest expressed in the infoset. (A WebIDL representation can be mapped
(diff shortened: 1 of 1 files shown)


mattgarrish:
I don't disagree with taking away the note designation, but I don't feel like this really belongs in the resource list section. I wonder if the bounds warrants its own subsection under s. 3 with a little more explanation, or if it belongs under "3.6 resources"?

iherman:
I think I agree. My preference would be to have a separate (short) section under s. 3.

iherman:
(Matt, I am still at TPAC, can you do it?)

mattgarrish:
Will give it a go. Was just switching source when you commented... :)

lrosenthol:
This looks good to me.  (there are some things we'll need to discuss further, but as far as this particular issues goes - I'm good!)

PR #355, opened 2018-10-29
Moved the canonicalization out of the top level

### LINKED ISSUE #349: Explain the purpose of the "canonical manifest"
The canonical manifest section appears without much in the way of introduction/explanation, which makes it confusing to understand. Yes, it's how you can get a json data structure from the manifest and the infoset rules, but why does this matter? Who is expected to create this manifest? When? In what context?

If the infoset is a non-language specific internal representation, then why define one process that creates a json representation of the infoset, in other words? Is it just intended as an example of how you might apply the rules for one kind of data representation?

An introduction that helps clear up its inclusion in the specification would really help.

iherman:
Did some editing to answer to the issues raised in #349. All changes are purely editorial.

Fix #349.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/355.html" title="Last updated on Oct 31, 2018, 9:39 AM GMT (a742882)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/355/b47ba5e...a742882.html" title="Last updated on Oct 31, 2018, 9:39 AM GMT (a742882)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3074,10 +3074,99 @@ partial dictionary WebPublicationManifest {
 		<section id="wp-lifecycle">
 			<h2>Web Publication Lifecycle</h2>
 
-			<p class="note">See the <a href="#app-lifecycle-diagrams">diagrams in the appendix</a> for a visual
-				representation of the lifecycle algorithm.</p>
+			<p class="note">
+				See the <a href="#app-lifecycle-diagrams">diagrams in the appendix</a> for a visual representation of the lifecycle algorithm.
+			</p>
+		
+			<section id="obtaining-manifest">
+				<h3>Obtaining a manifest</h3>
+
+				<p> The <dfn data-lt="obtaining the manifest|obtaining a manifest|obtains a manifest">steps for
+						obtaining a manifest</dfn>, starting from the <a>primary entry page</a>, are given by the
+					following algorithm. The algorithm, if successful, returns a <a>processed manifest</a>; otherwise,
+					it terminates prematurely and returns nothing. In the case of nothing being returned, the user agent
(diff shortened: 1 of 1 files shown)


mattgarrish:
A couple of comments:

- it still doesn't seem intuitive to introduce the steps for generating the canonical manifest before any mention is made of it in the specification; that's what threw me the first time. It would be better to have it follow the section on obtaining the manifest, like the section on processing does, than try to link forward to where it matters.
- slightly unrelated to this issue, but why is a json object parsed from text in step 7, but the canonical manifest is created from text again in step 9? Similarly, it looks like creating a canonical manifest results in a json object that could be used for processing, but that step also begins with parsing text. After obtaining the json in step 7, aren't you continuing to manipulate that object throughout the subsequent steps?

mattgarrish:
Also, rather than calling the section "Canonical Manifest" it might be helpful to provide a name that matches the lifecycle nature of the section, so something like "Generating a Canonical Manifest".

iherman:
You are right on all accounts:-)

I have changed the order as you suggested, changed the title, and also modified the algorithm. Parsing is only done once, and the result of parsing (`json`) is input to the canonicalization step. There is also an additional remark in that algorithm that the input is a json object representing the manifest.

Thx:-)

mattgarrish:
The processing section looks like it still needs some tweaks to accept a json object that is the canonical manifest rather than parsing text, but otherwise looks better now.

iherman:
I will do that tomorrow morning and then merge. ok?

mattgarrish:
sgtm

PR #358, opened 2018-11-01
Made some minor changes

iherman:
(This is a PR for a non-master branch!)

- I made changes in the new section on authored vs. canonical manifest
- in the case of embedded manifest's base URL I have added reference to the pending issues, and also removed the reference to xml:base (it has been removed from the latest HTML versions)

### DIFF
--- a/index.html
+++ b/index.html
@@ -237,7 +237,7 @@
 					<p>The manifest is expressed in one of two forms depending on the state of the Web Publication:</p>
 
 					<dl>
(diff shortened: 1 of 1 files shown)


iherman:
@mattgarrish I like what you have! To avoid a branch mess, this should probably be PR-d and then merged into master soon...

mattgarrish:
Yes, I committed a first pass which was just a basic search to remove instances of "infoset". I started a second pass to read through and make sure everything still made sense. I should have that done later today at which point I'll open a PR.

PR #359, opened 2018-11-01
Remove infoset

mattgarrish:
This PR removes the "infoset" concept and replaces it with more emphasis on the authored v. canonical manifests.

Prior to this, we had infoset and manifest subsections for almost every property, whereas now the information has been merged and, in most cases, the manifest expressions moved up after the introductory paragraph.

As this change required a significant reread of the specification, I've also made a number of minor editorial tweaks along the way. These are generally insignificant, with the exception of the language and direction section. As I read through, I noticed that most of the global language info was specified before the subsection containing the property definitions (plus at least one para that looked like it belonged with the overrides). @iherman , please have a look and see if my shuffling still makes sense to you.

Otherwise, I think there's a diagram or two that will need updating, but I didn't try and fix those.

### DIFF
--- a/index.html
+++ b/index.html
@@ -61,9 +61,9 @@
 	<body>
 		<section id="abstract">
 			<p>This specification defines a collection of information that describes the structure of Web Publications
-				so that user agents can provide user experiences tailored to reading publications, such as sequential
-				navigation and offline reading. This information includes the default reading order, a list of
-				resources, and publication-wide metadata.</p>
+				so that user agents can provide user experiences specially tailored to reading publications, such as
+				sequential navigation and offline reading. This information includes the default reading order, a list
+				of resources, and publication-wide metadata.</p>
 		</section>
 		<section id="sotd">
 			<p>This draft provides a draft version of a Web Publication. Many details are under active consideration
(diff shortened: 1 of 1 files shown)


mattgarrish:
Not sure why no preview option is showing up. Only thing I can think is because the diff is too large. Respec doesn't show me any issues.

iherman:
I do not know why preview does not work either. 

This URI seems to work to display the new version:

https://tinyurl.com/y9ds858k

iherman:
I have committed the document with some minor changes.

mattgarrish:
Yes, I had the same thought last night, but even if authored=canonical, you don't actually get canonical until the user agent has processed the authored manifest, correct? That was the distinction I was trying to draw, between the one that someone/something has produced versus the post-processed.

Would it help to add some more clarification to that effect?

iherman:
Well, if authored=canonical then the canonicalization is an identify operation... But yes, it will still happen, in this sense it is not 100% the same.

I think adding a somewhat more verbose clarification would indeed be the best, rather than producing a concise and fully precise text...

mattgarrish:
I added another para, but feel free to mod directly if you have other ideas.

iherman:
Ok, let us see what others think. It works for me...

mattgarrish:
Here's an alternative preview that doesn't break when you click internal links:
https://cdn.staticaly.com/gh/w3c/wpub/remove-infoset/index.html

mattgarrish:
Should we go ahead and merge this so we don't block the other open issues? (It's mostly shuffling, so if it can benefit from further clarification that shouldn't impact on what exists now.)

PR #362, opened 2018-11-06
Turned all occurences of PublicationLink to Linkedresource

### LINKED ISSUE #235: Use LinkRole instead of PublicationLink
(Extracting a discussion on #232 into a separate Issue.)

Schema.org has a (currently "pending") type [LinkRole](https://pending.schema.org/LinkRole) which may be a good alternative to the (publication specific) [PublicationLink](https://w3c.github.io/wpub/#publicationLink). Maybe worth considering using the schema.org type.

---

Ref: https://github.com/w3c/wpub/pull/232#pullrequestreview-129576214, https://github.com/w3c/wpub/pull/232#issuecomment-398077232, https://github.com/w3c/wpub/pull/232#issuecomment-398079325

### LINKED ISSUE #356: Is PublicationLink a misnomer?
Using the right terminology is important, because it also suggests some sort of a "mental model" of what we are doing. In this sense, I would suggest that the term "PublicationLink" is not the right term, mainly with the additional features proposed (for good reasons!) for audio or video books. I would suggest we would use the term `PublicationResource` (we can bike shed on a better term) and some of the constituent terms may also have to be changed.

If we agree, we should also close #235 ("Using LinkRole instead of PublicationLink") as being overtaken by events.

@BigBlueHat @HadrienGardeur @laudrain @TzviyaSiegman @wareid @llemeurfr

iherman:
See discussion in #356, this is the PR embodying the outcome.

Fix #356
Fix #235

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/362.html" title="Last updated on Nov 6, 2018, 6:09 AM GMT (523ddee)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/362/8e62b59...523ddee.html" title="Last updated on Nov 6, 2018, 6:09 AM GMT (523ddee)">Diff</a>


HadrienGardeur:
This looks fine overall, but in the WebIDL definition of `LinkedResource` I see that `description` is a `LocalizableString` instead of a `sequence<LocalizableString>`.

Is that on purpose?

iherman:
@HadrienGardeur that is already the case for the current official draft, ie, it is not specific to this PR.

To be honest, I do not remember why it was set to a single value instead of a sequence, but it is consistent with the [definition](https://w3c.github.io/wpub/#publicationLink). I would propose to de-associate this issue from this PR and raise it separately if you think it is a mistake.

HadrienGardeur:
@iherman that's correct, I only noticed it today because I was specifically reviewing the section related to `LinkedResource`. I'll open a separate issue.

iherman:
This is, essentially, an editorial PR (simple renaming), so I will merge and close this...

PR #370, opened 2018-11-27
Start work on WPUB Explainer document

dauwhe:
I've started an explainer document for the WPUB spec. Having an [explainer](https://github.com/w3ctag/w3ctag.github.io/blob/master/explainers.md) is a best practice for creating specs. The goal is to describe what we're doing and why. In our case we hope this will help us get useful feedback from the W3C, the TAG, and the developer community. 

This is obviously in an early stage, and would benefit from input from all of us.

### DIFF
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -18,8 +18,8 @@ Thus the goal of web publications is to make these features ("affordances") avai
  
 - Provide a mechanism for defining a collection of web resources as a publication
 
- - The resources may be images (comics/manga) or audio files (audiobooks)
- 
+ - The resources may be images (comics/manga), audio files (audiobooks), or even data files which would not be rendered.
+  
 - Provide a mechanism for ascribing descriptive metadata to a collection of web resources.
 
 - Provide a mechanism for defining an ordering of the resources in a publication
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -24,7 +24,7 @@ Thus the goal of web publications is to make these features ("affordances") avai
 
 - Provide a mechanism for defining an ordering of the resources in a publication
 
-- Describe the affordances needed for reading publications on the web
+- Describe the affordances needed for reading publications on the web. Note that publications may take many forms, from journals to educational texts to novels. 
 
  
  
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -142,7 +142,7 @@ The key questions are [1] identifying the "bounds" of the publication, [2] defin
 
 ```
 
-The web publications spec has essentially adopted EPUB's approach, with an explicit list of resources. Anything that is not part of the `resources` or `readingOrder` manifest members is considered to be outside the web publication. 
+The web publications spec has essentially adopted EPUB's approach, with an explicit list of resources. Anything that is not part of the `resources` or `readingOrder` manifest members is considered to be outside the web publication. The group considered things like `scope` and URL patterns, but felt that an explicit list was simpler. 
 
 ### 2. Sequence of primary resources
 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -11,7 +11,7 @@ Two things make web publications different from the “ordinary” web we know a
 
 More importantly, users have a set of expectations about how such content should be presented in order to make it easy to read and understand. Users need to personalize the presentation, using the font and font size that make it easiest for them to read. They want it to be easy to go to the next chapter without interrupting the reading experience by hunting for a link to click. They might need a high- or low-contrast version of the content. They want to read while offline.
 
-Thus the goal of web publications is to make these features ("affordances") available to readers. We hope that someday browsers will provide these features, but in the meantime we have to use scripting. That's OK, because we'll learn a lot from trying, and that will help the spec get better.
+Thus the goal of web publications is to make these features ("affordances") available to readers. We hope that someday browsers will provide these features, but in the meantime we have to use scripting and polyfills. That's OK, because we'll learn a lot from trying, and that will help the spec get better.
 
 ## Goals
 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -182,10 +182,6 @@ The Web Publication Manifest appears to be very similar to the Web Application M
 3. Web publications are fundamentally different from web apps, as the goal is for the user agent to provide the user interface. 
 
 
-But note that the [TAG has spoken](https://github.com/w3c/wpub/issues/32#issuecomment-362273649) in favor of using WAM. 
-
-
-
 ## The User Experience
 
 Reading something that takes a day or a week rather than a few minutes influences what sort of user experience works for the reader. We have a formal [use cases and requirements document](https://w3c.github.io/dpub-pwp-ucr/) in addition to lots of actual experience from ebook reading systems. Browsers have implemented "reading modes" to address some of these requirements. 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -177,7 +177,7 @@ The Web Publication Manifest appears to be very similar to the Web Application M
 
 1. WP use cases are orthogonal to those of WAM. Nothing is stopping a creator of a web publication from also using a web application manifest, if the publication author desires for the publication to be installable, etc. 
 
-2. The web application manifest spec is not designed to be extensible. 
+2. There were concerns about the extensibility of web application manifest, especially with regards to the processing of new members. 
 
 3. Web publications are fundamentally different from web apps, as the goal is for the user agent to provide the user interface. 
 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -218,6 +218,10 @@ Reading something that takes a day or a week rather than a few minutes influence
 
 - it should be possible to search the entire publication
 
+## Unanswered questions
+
+1. Addressability. A publication needs a permanent URL, and it should be possible to craft a URL to point anywhere within a publication. 
+
 ## Acknowledgments
 
-The initial draft was written by Dave Cramer. 
+The initial draft was written by Dave Cramer. Thanks to Ivan Herman and Tzviya Siegman for helpful feedback. 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -161,7 +161,7 @@ Once again, web publications use an explicit list outside of the content itself,
 
 - With transclusion, one could define metadata in the parent file to apply to the whole.
 
-Web publication metadata is expressed in the manifest, using a vocabulary largely taken from schema.org. 
+Web publication metadata is expressed in the manifest using the schema.org vocabulary, with some extensions. The metadata is serialized as JSON-LD, one of the three syntaxes accepted by schema.org (the others being microdata and RDFa). JSON-LD has also the advantage of being easy to use for both client and server side processing.
 
 
 
--- a/EXPLAINER.md
+++ b/EXPLAINER.md
@@ -38,7 +38,7 @@ Thus the goal of web publications is to make these features ("affordances") avai
  
 ## Basic design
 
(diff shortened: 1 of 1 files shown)


EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+The Web Publication Manifest appears to be very similar to the Web Application Manifest. Both are JSON files that are linked to from HTML, and provide metadata about a composite resource. Several [arguments](https://github.com/w3c/wpub/wiki/Options-for-Processing-a-Manifest) [have](https://github.com/w3c/wpub/issues/32) been made against using WAM.
+
+1. WP use cases are orthogonal to those of WAM. Nothing is stopping a creator of a web publication from also using a web application manifest, if the publication author desires for the publication to be installable, etc. 
+
+2. The web application manifest spec is not designed to be extensible. 
+
+3. Web publications are fundamentally different from web apps, as the goal is for the user agent to provide the user interface. 
+
+
+But note that the [TAG has spoken](https://github.com/w3c/wpub/issues/32#issuecomment-362273649) in favor of using WAM.
```
iherman: I think that TAG comment was at such an early stage of our work, that I am not sure it is still relevant. We did spend a huge amount of time on this issue, and we made the decision of going our own way; I think it is better if the TAG looks at it with a fresh eye. 

Bottom line: I do not think this reference should stay in the explainer.

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+
+Two things make web publications different from the “ordinary” web we know and love. First, a web publication may consist of multiple resources that form a logical whole. Moby-Dick might consist of 136 HTML files in a specified order, but it’s still a single work. So searching should search all 136 chapters.
+
+More importantly, users have a set of expectations about how such content should be presented in order to make it easy to read and understand. Users need to personalize the presentation, using the font and font size that make it easiest for them to read. They want it to be easy to go to the next chapter without interrupting the reading experience by hunting for a link to click. They might need a high- or low-contrast version of the content. They want to read while offline.
+
+Thus the goal of web publications is to make these features ("affordances") available to readers. We hope that someday browsers will provide these features, but in the meantime we have to use scripting. That's OK, because we'll learn a lot from trying, and that will help the spec get better.
```
iherman: We may be more explicit than just 'scripting': we can say that we may see polyfills, full blown applications, etc, coming up, and the goal of the work is to make these choices possible.

Maybe it is worth referring to an MVP in this respect?

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Goals
+
+ 
+- Provide a mechanism for defining a collection of web resources as a publication
+
+ - The resources may be images (comics/manga) or audio files (audiobooks)
```
iherman: Or data files (important, because these are not directly interpretable by browsers, for example).

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+      <iframe src="c002.html">
+      <iframe src="c003.html">
+    </web-publication>
+  </body>
+</html>
+
+```
+
+The web publications spec has essentially adopted EPUB's approach, with an explicit list of resources. Anything that is not part of the `resources` or `readingOrder` manifest members is considered to be outside the web publication. 
+
```
iherman: Note that the group did consider the `scope` feature or other features like URI patterns, and I believe the feeling was that we would stick to a simple solution for this version and we may revisit this for later versions. (Again to avoid unnecessary discussions with the readers...)

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+- HTML can express ordering of resources via `rel=prev` and `rel=next`, but these do not have UI support from the browsers (with notably rare exceptions).
+
+- HTML can also express an ordered list of links with the `nav` element. 
+
+- Transclusion can also define an ordering of the transcluded components. 
+
+- EPUB uses the `spine` element in the XML package file to determine a primary reading order.
+
+Once again, web publications use an explicit list outside of the content itself, via the `readingOrder` member of the manifest. 
+
```
iherman: Do we want to also say that WP also has the notion of a TOC for a more detailed set of references (which may or may not be the same as the resources) and this is based on HTML structures (eg, `nav`)?

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+- EPUB includes metadata in the XML package file
+
+- There are many ways of embedding metadata in individual HTML files
+
+- Metadata about a web application can be expressed in the web application manifest file. 
+
+- With transclusion, one could define metadata in the parent file to apply to the whole.
+
+Web publication metadata is expressed in the manifest, using a vocabulary largely taken from schema.org. 
+
```
iherman: re-emphasize: and a syntax favoured by schema.org.

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+
+
+### Relationship to Web Application Manifest
+
+The Web Publication Manifest appears to be very similar to the Web Application Manifest. Both are JSON files that are linked to from HTML, and provide metadata about a composite resource. Several [arguments](https://github.com/w3c/wpub/wiki/Options-for-Processing-a-Manifest) [have](https://github.com/w3c/wpub/issues/32) been made against using WAM.
+
+1. WP use cases are orthogonal to those of WAM. Nothing is stopping a creator of a web publication from also using a web application manifest, if the publication author desires for the publication to be installable, etc. 
+
+2. The web application manifest spec is not designed to be extensible.
```
iherman: That may not be 100% true, because they do talk about extension points, but it is very unclear how that would be used for an extension of processing.

EXPLAINER.md
dauwhe: I'm concerned that we would be seen as ignoring their feedback (which, in fact, is true). I see some value in acknowledging that we've at least aware of their concerns.

EXPLAINER.md
dauwhe: I can add something. But there's nothing on the web today stopping us from hosting data files. We aren't envisioning that these data files will require new web functionality, and my understanding is that we don't envision having an ordered sequence of data files as our reading order?

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+ - DRM
+ 
+## Basic design
+
+A web publication must have an <i>entry page</i>, which the HTML document returned by the URL of the publication. This page must have either a link to the manifest (`<link rel="publication" href="manifest.json">`, or an [embedded](https://github.com/w3c/wpub/issues/327) manifest. 
+
+A “manifest” is a list of the passengers or cargo on a ship. For web publications, a manifest lists the constituents of the publication—all the HTML files, stylesheets, images, scripts, etc.—needed to create the whole. It further describes the sequence of primary resources, so that we know that chapter-02.html comes after chapter-01.html. In EPUB we called this list the "spine"; for web publications it's now the `readingOrder`. 
+
+The manifest is also the natural location for metadata that applies to the whole publication, rather than just one of the constituents. The metadata vocabulary is based on schema.org; the entire manifest is serialized as JSON-LD. 
+
```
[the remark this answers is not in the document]
dauwhe: It is perhaps notable that we chose only a single serialization of schema.org :)

EXPLAINER.md
iherman: I understand, but I am still worried...

EXPLAINER.md
iherman: All true. But there are many people working on the web who needs a reminder that data on the web are genuinely part of the picture...:-)

EXPLAINER.md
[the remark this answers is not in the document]
iherman: True, but I defy you to put all the content of the manifest into RDFa, let alone microdata... Been there, done that!

EXPLAINER.md
dauwhe: Does 9534098 address your concerns?

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+ - The resources may be images (comics/manga) or audio files (audiobooks)
+ 
+- Provide a mechanism for ascribing descriptive metadata to a collection of web resources.
+
+- Provide a mechanism for defining an ordering of the resources in a publication
+
+- Describe the affordances needed for reading publications on the web
+
+
```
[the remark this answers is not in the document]
dauwhe: Does 4754bb6 help?

EXPLAINER.md
dauwhe: Deleted in 
c149443

EXPLAINER.md
dauwhe: We mention the TOC in the section on navigation. I'm adding a line to the manifest section saying we can refer to an HTML TOC there. See b789693.

EXPLAINER.md
dauwhe: See 99ba6f6

EXPLAINER.md
```diff
@@ -0,0 +1,227 @@
...
+Note that `readingOrder` defines the sequence of primary resources that form the publication. `resources` enumerates all the other resources that are required to render the publication. 
+
+## Design choices
+
+The unanswered question:
+
+> is the goal of this group to make a new packaging format for specialist book reading software and devices, or is it to obtain first class support for missing book-related features in the web platform as a whole?
+
+
```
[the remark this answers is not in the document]
dauwhe: I think the quote from Andrew Betts is referring to more than just packaging, but the whole direction of the working group. We've certainly spent much of our time on a high-level data model, rather than identifying low-level primitives that would allow us to more easily build publications on the web. For good or for bad, what we are actually building can most accurately be described as unpackaged EPUB with the OPF serialized as JSON-LD.  

Packaging should and will have a great influence on what happens with WPUB. For example, web packages right now explicitly depend on a manifest, and actually refer to ours as well as WAM. Having information in our manifest that would enable packaging would be very helpful. I hope we avoid a situation where most every publication needs a separate web publication manifest, a web application manifest, and a web packaging manifest.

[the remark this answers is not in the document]
iherman: Dave, I do not think we disagree; this shows, however, that just having this question thrown into the explainer without further explanation may not be the right thing to do. That is what motivated my original comment.

EXPLAINER.md
dauwhe: Do we need to mention JSON-LD again here?

iherman: I am just defensive... Do you think it is an overkill?

iherman: I have heard too many times 'why json-ld?'

dauwhe: Ivan, could you write a few sentences on "Why JSON-LD"? I figure you are far more qualified to do this than I am :)

EXPLAINER.md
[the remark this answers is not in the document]
dauwhe: Ivan, I've moved that question to the "unanswered questions" section, included more of the quote, and tried to provide some context in 312cd04.

EXPLAINER.md
dauwhe: I've added a mention of polyfills. 

I'm a little wary of mentioning the concept of a MVP. Perhaps it's better to start by explaining the whole picture without prioritizing features.

EXPLAINER.md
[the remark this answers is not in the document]
iherman: I am still a bit wary with this question, but I certainly prefer as part of the unanswered questions. Let us do this for now, and see what others comment...

EXPLAINER.md
iherman: How does that sound?

> Web publication metadata is expressed in the manifest using the schema.org vocabulary, with some extensions when necessary. The serialization syntax used for the manifest is JSON-LD: among the three syntax's accepted by schema.org processors (the others being microdata and RDFa) this one has also the advantage of being easy to use for various client or server side processing. 

(You may make it more fluid in English...)

dauwhe: Thanks Ivan! Incorporated in 4f22d57.

iherman:
LGDM!

Issue #372, opened 2018-12-04
Write an audiobook explainer


dauwhe:
We should write an [explainer document](https://github.com/w3ctag/w3ctag.github.io/blob/master/explainers.md) as we design audiobooks, and share it with the TAG as early as possible.

wareid:
https://github.com/wareid/audiopub-explainer

wareid:
Adding propose closing since we have a PR in place to add the explainer :).

dauwhe:
And we now have an audiobook explainer, thanks to @wareid !

PR #373, opened 2018-12-04
Create AUDIOBOOKS_EXPLAINER.md

wareid:
Adding the Audiobooks Explainer to the WP repo. This is an initial draft and I look forward to feedback from the group. I would like to use this for distribution to audio publishers and interested parties. 

Thanks to @dauwhe's issue #372 for reminding me to add this!

### DIFF
--- a/AUDIOBOOKS_EXPLAINER.md
+++ b/AUDIOBOOKS_EXPLAINER.md
@@ -7,11 +7,14 @@ Audiobooks have grown massively in popularity in the last few years despite bein
 * Improved mobile networks (3G, 4G LTE)
 * Improved mobile storage capacity
 * Widespread use of wifi
+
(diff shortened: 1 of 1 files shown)


iherman:
I have an organizational comment (and I should have spotted that when @dauwhe did the overall explainer): the way we go is that we might have a number of specialized publication types and each would deserve its own explainer. I think it may be better if all explainers were collected in a separate folder (called, say, "Explainers"), and possibly with its own index.md file as a table of content.

@TzviyaSiegman, w.d.y.t?

TzviyaSiegman:
@iherman I think this will tidy up the repo. Good idea.

TzviyaSiegman:
@iherman I think this will tidy up the repo. Good idea.

PR #383, opened 2018-12-10
add more detail about other options to audiobook explainer

dauwhe:


BigBlueHat:
@dauwhe maybe link to https://en.wikipedia.org/wiki/MPEG-4_Part_14 for M4B? Otherwise, looks great to me. Thanks for writing this up!

Issue #386, opened 2018-12-13
Must the primary entry page be part of the web publication?


dauwhe:
So... the primary entry page doesn't need to be part of `readingOrder`:

> It is not required that the primary entry page be included in the default reading order, nor that it be the first document listed when it is included. This specification leaves the exact nature of the document intentionally underspecified to provide flexibility for different approaches (e.g., the primary entry page could be a marketing document for the Web Publication instead of a specific page of content). If a default reading order is not provided, however, the primary entry page will be used as the default entry.

Does the primary entry page need to be part of the web publication? What if the manifest is embedded in the primary entry page, and not linked from any resource of the publication? Are we allowing all the information that defines a web publication to be entirely outside the bounds of the publication? If such a web publication were to be packaged, how would the content of the manifest be included? 

How does this affect the initialization of a web publication? A user navigates to the primary entry page. A user agent must load at least enough of the page to find the manifest, and then determine if the entry page is part of the publication. Should the user agent initiate publication mode for the entry page, even though it's not part of the publication? Should it automatically navigate to the first item in `readingOrder`? Or does the user have to initiate that navigation, as well as opting in to publication mode?

I don't suppose it matters, but it seems odd that the the resource returned by the publication's URL might not be part of the publication.

iherman:
I do not remember the history of this, to be honest. I _guess_ the idea was that the PEP may only contain a TOC, for example, that is taken care of by the User Agent, hence it is not necessarily in the reading order.

But I agree that it should be part _either_ of the reading order _or_ the resources. We have defined the "borders" of the publication to be derived from these two, and I think it would be a reasonable requirement to say that the PEP is in one of these two.

(I suspect that this fuzziness around the PEP came from the time when there was also a fuzziness about the bounds of a publication, actually.)

mattgarrish:
> I _guess_ the idea was that the PEP may only contain a TOC, for example, that is taken care of by the User Agent, hence it is not necessarily in the reading order.

As I recall, there was a question of whether the PEP would be a marketing page or page from which you could buy the publication, for example, but not actually be a resource that you would read as part of the publication.

I agree it's confusing that it doesn't have to be a resource, but should we be requiring it as a resource if there's no intention of rendering it?

iherman:
> I agree it's confusing that it doesn't have to be a resource, but should we be requiring it as a resource if there's no intention of rendering it?

[...]

PR #389, opened 2018-12-19
Clarify the primary entry page

mattgarrish:
This PR addresses two of the concerns raised in issue #386:

- more clearly state that the PEP is a required resource
- remove the parenthetical about what content the PEP might contain

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/389.html" title="Last updated on Dec 19, 2018, 2:27 PM UTC (0f1cda3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/389/ae7c8b2...0f1cda3.html" title="Last updated on Dec 19, 2018, 2:27 PM UTC (0f1cda3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -745,7 +745,7 @@ partial dictionary WebPublicationManifest {
 				<p>The <dfn>primary entry page</dfn> represents the preferred starting <a href="#wp-resources"
 						>resource</a> for a Web Publication and enables discovery of its <a>manifest</a>. It is the
 					resource that is returned when accessing the Web Publication's <a>address</a>, and MUST be included
-					either in the <a>default reading order</a> or <a>resource list</a>.</p>
+					in either the <a>default reading order</a> or the <a>resource list</a>.</p>
 
 				<p>Although any resource can link to the Web Publication manifest, the primary entry page typically
 					introduces the publication and provides access to the content. It might contain all the content, in
--- a/index.html
+++ b/index.html
@@ -742,10 +742,10 @@ partial dictionary WebPublicationManifest {
 			<section id="wp-primary-entry-page">
 				<h3>Primary Entry Page</h3>
 
-				<p>Every Web Publication MUST include a <dfn>primary entry page</dfn>, which is the resource that MUST
-					be returned when accessing the Web Publication's <a>address</a>. The primary entry page represents
-					the preferred starting <a href="#wp-resources">resource</a> for the Web Publication, and enables
-					discovery of the <a>manifest</a>.</p>
+				<p>The <dfn>primary entry page</dfn> represents the preferred starting <a href="#wp-resources"
+						>resource</a> for a Web Publication and enables discovery of its <a>manifest</a>. It is the
+					resource that is returned when accessing the Web Publication's <a>address</a>, and MUST be included
+					either in the <a>default reading order</a> or <a>resource list</a>.</p>
 
 				<p>Although any resource can link to the Web Publication manifest, the primary entry page typically
 					introduces the publication and provides access to the content. It might contain all the content, in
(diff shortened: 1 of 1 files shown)


HadrienGardeur:
It's a fallback, so we shouldn't be talking about a "default entry" IMO.

It's meant to make things easier for "single resource in the reading order" publications.

mattgarrish:
> It's a fallback, so we shouldn't be talking about a "default entry" IMO.

Right, it's just trying to capture that it gets inserted when creating the canonical manifest if no reading order is specified. I've removed the default entry aspect and linked it to the canonical manifest step.

iherman:
The latest commit from @mattgarrish is in line with the text in [4.5.1](https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/389.html#default-reading-order):

> The default reading order is specified directly in the manifest, but MAY be omitted when it only consists of the primary entry page . When the default reading order is absent, user agents MUST include an entry for the primary entry page when compiling the canonical manifest. 

+1 thus

iherman:
Let us note, however, that this PR does _not_ solve issue #386. It does not say that the PEP should be within the bounds of the publication. (Which is all right for this PR, it is up to us to drive #386 to a conclusion...)

mattgarrish:
> It does not say that the PEP should be within the bounds of the publication.

Isn't that required now? If the WP MUST have a PEP, then it has to be specified either in the resources or the default reading order, which puts it within the bounds.

Would it help to add a parenthetical to this effect after the first MUST, like:

> Every Web Publication MUST include a <dfn>primary entry page</dfn> (i.e., this resource has to be listed in the default reading order or resources),

iherman:
>> It does not say that the PEP should be within the bounds of the publication.
>
> Isn't that required now? If the WP MUST have a PEP, then it has to be specified either in the resources or the default reading order, which puts it within the bounds.

Where does it say that? I was looking at 3.4, and I did not see this stated. It just says it is a "resource" (in its general form), but that may mean it is part of the "links" array...

So yes, I believe this should be explicitly stated.

mattgarrish:
I modified the first paragraph of 3.4 to use MUSTs, so it now reads:

> Every Web Publication MUST include a primary entry page, which is the resource that MUST be returned when accessing the Web Publication's address. The primary entry page represents the preferred starting resource for the Web Publication, and enables discovery of the manifest.

Or that's what you should be seeing if the preview isn't doing something weird.

If we want it clearer, I can add a parenthetical like I mentioned above.

iherman:
@mattgarrish that is what I see indeed. And it does not say that, formally, it must be part of either the reading order _or_ the resources array in the manifest, and that is what defines the boundaries...

I think adding the parenthetical remark is actually necessary.

mattgarrish:
I've rewritten the paragraph, as we're starting to repeat statements elsewhere in the specification. In the immediately preceding Resources section, for example, we say:

> A Web Publication MUST include at least one HTML document [html]—the primary entry page.

Instead of repeating that again, I've moved the definition of what it is to the start and made the only MUST that it has to be in one of those lists. The address section also states that it is equivalent of the PEP, so I don't think we need to normatively say again that is the expectation. For quick reference, it's now:

> The primary entry page represents the preferred starting resource for a Web Publication and enables discovery of its manifest. It is the resource that is returned when accessing the Web Publication's address, and MUST be included either in the default reading order or resource list.

iherman:
As far as I am concerned it is a go!

toshiakikoike:
Is it not necessary to define the primary entry page in the manifest?

https://w3c.github.io/wpub/#simple-book
C.1. Simple Book
There are no definition of primary entry page in either "readingOrder" or "resources".

Or is the HTML pointed to by "url" (by default "./index.html") treated as a primary entry page?
In other words, in the "Simple Book" sample,
is "https://publisher.example.org/mobydick/index.html" treated as a primary entry page?

mattgarrish:
> Or is the HTML pointed to by "url" ... treated as a primary entry page?

Yes.

mattgarrish:
To respond a little more completely (was travelling earlier), the `url` property has to reference the primary entry page. As noted in the recent tag review, there are a couple of paragraphs that describe what the address should reference, but those are outdated and needed to be removed in this pull request. They will be removed in a pending update.

PR #392, opened 2019-01-09
suggested mods to design section

mattgarrish:
Sorry I haven't had time to look this over sooner, but generally I'm fine with the document. The one section that confused me, though, was the one on design. It wasn't entirely clear what the reader is supposed to get from that section, or what questions it's asking. I also find leading each subsection with bullets and then getting to what the specification currently adopts isn't ideal.

But instead of griping, I did some rewriting to what I think the section is trying to get at. Feel free to use/discard whatever see fit.

### DIFF
--- a/explainers/wpub-explainer.md
+++ b/explainers/wpub-explainer.md
@@ -107,18 +107,24 @@ Note how we used `rel=contents` to identify that a particular resource is an HTM
 ## Design choices
 
 
-The key questions are [1] identifying the "bounds" of the publication, [2] defining the ordering of the primary resources, and [3] figuring out how to express metadata for the publication as a whole. 
+The design of web publications has caused perhaps the greatest amount of discussion within the group, and centers around three key areas:
+
+1. how best to identify the "bounds" of the publication;
+2. how best to define the ordering of the primary resources; and
+3. how best to express metadata for the publication as a whole.
+
+This section outlines the current approaches that have been adopted in each of these areas, and lists alternative practices that have been discussed and/or influenced the decision-making process.
 
 
-### 1. What's part of the publication?
+### 1. What is part of the publication?
 
-- The [Web Application Manifest spec](https://w3c.github.io/manifest/) uses `scope` to define the extent of a web application. 
+The web publications working group has chosen to use an explicit list of resources in the manifest to identify the bounds, similar to how EPUB approaches the problem. Anything that is not part of the `resources` or `readingOrder` manifest members is considered to be outside the web publication. The group has also considered the patterns described below, but currently believes an explicit list is the simpler choice.
 
-- Every component of an EPUB publication must be listed in a `manifest` element in an XML package file.
+- Using a `scope`, as was done in the [Web Application Manifest spec](https://w3c.github.io/manifest/).
 
-- Web sites do not explicitly define their boundaries.
+- Not explicitly define the boundaries, similar to how web sites function.
 
-- XML sometimes uses the idea of *transclusion*, where the contents of other documents can be incorporated into a parent document based on hypertext references. This concept can also be realized in HTML using iframes, HTML imports, custom elements, etc. The working group has shown no interest in this approach. 
+- Using the notion of *transclusion*, where the contents of other documents can be incorporated into a parent document based on hypertext references (e.g., through iframes, HTML imports, custom elements, etc.). For example:
 
 ```html
 <html>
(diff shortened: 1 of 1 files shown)


explainers/wpub-explainer.md
```diff
@@ -136,33 +142,21 @@ The key questions are [1] identifying the "bounds" of the publication, [2] defin
...
-
-- EPUB includes metadata in the XML package file
-
-
-- Metadata about a web application can be expressed in the web application manifest file. 
-
-- With transclusion, one could define metadata in the parent file to apply to the whole.
-
-Web publication metadata is expressed in the manifest using the schema.org vocabulary, with some extensions. The metadata is serialized as JSON-LD, one of the three syntaxes accepted by schema.org (the others being microdata and RDFa). JSON-LD has also the advantage of being easy to use for both client and server side processing.
+Web publication metadata is expressed in the manifest using the schema.org vocabulary, with some extensions. The metadata is serialized as JSON-LD, one of the three syntaxes accepted by schema.org (the others being microdata and RDFa). JSON-LD has also the advantage of being easy to use for both client and server side processing. As previously, this method is similar to the EPUB approach.
```
iherman: "As previously, this method is similar to the EPUB approach." I am not sure this is true in this case. There are, obviously, similarities, but the fact of using schema.org and abiding to an external, standard syntax for metadata is, I believe, a significant change compared to EPUB. It is line with the general approach of alignment of publication technologies with what is happening on the Web at large (something worth mentioning, b.t.w.)

explainers/wpub-explainer.md
```diff
@@ -136,33 +142,21 @@ The key questions are [1] identifying the "bounds" of the publication, [2] defin
...
 
+Alternatively, it is possible to embed metadata directly in the header and/or content of HTML files, which could include the parent of a set of transcluded files so that it applies to all the sub-documents.
```
iherman: Maybe worth emphasizing at this point that, at this time, most of the schema.org processors (if not all) rely on the embedded JSON-LD approach rather than the separate file. The reason this is worth emphasizing is that the embedding approach was rejected for the Web Application Manifest (after some discussions), i.e., our decision is in contrast to what they decided to do.

explainers/wpub-explainer.md
```diff
@@ -107,18 +107,24 @@ Note how we used `rel=contents` to identify that a particular resource is an HTM
...
-### 1. What's part of the publication?
+This section outlines the current approaches that have been adopted in each of these areas, and lists alternative practices that have been discussed and/or influenced the decision-making process.
 
-- The [Web Application Manifest spec](https://w3c.github.io/manifest/) uses `scope` to define the extent of a web application. 
 
-- Every component of an EPUB publication must be listed in a `manifest` element in an XML package file.
+### 1. What is part of the publication?
 
-- Web sites do not explicitly define their boundaries.
+The web publications working group has chosen to use an explicit list of resources in the manifest to identify the bounds, similar to how EPUB approaches the problem. Anything that is not part of the `resources` or `readingOrder` manifest members is considered to be outside the web publication. The group has also considered the patterns described below, but currently believes an explicit list is the simpler choice.
```
iherman: Maybe it is worth saying that the scope or transclusion approaches (but I think we also considered the usage of URI patterns, b.t.w.) may come back for later versions, and the decision for the simpler choice was also a pragmatic to ensure an easy transition for the publ. community.

explainers/wpub-explainer.md
mattgarrish: Sure, but I was just trying to align these subsections with the statement later on that the manifest to date basically represents a translation of the package document to the web. If we make that statement, we should point out how it is true where we discuss these design decisions. I can expand on that sentence, though, to put it in relation to both epub and the web at large.

iherman: Yeah... saying that it is a 'translation' may be misleading. I would probably prefer 'inspired by', which is somewhat weaker (and, I think, closer to reality). I missed this at the first review cycle...

And adding some words may indeed be helpful...

mattgarrish:
I've pushed new some changes to the document.

Per your comments, @iherman, I reworded the mentions of epub to indicate only that the structures serve as precedent for the choices. I also took the section on the relationship to WAM and moved it to the unanswered architecture question. The cited comment and the explanation of why WAM wasn't chosen seem to go hand-in-hand.

I also tempered a couple of statements, as the explainer really shouldn't lead readers to conclusions about the (in)correctness of what we've done. It's not that I don't agree with the statements in some cases, but the document should probably be as neutral as possible in fairness to everyone. Let's allow the reviewers to determine the fitness of the choices.

PR #393, opened 2019-01-17
Major changes to WP explainer

dauwhe:
I've attempted to rewrite the explainer for a world where we are just defining the manifest, as discussed in [Ivan's gist](https://gist.github.com/iherman/1311da6cde8e564eb44036c7caf863d1#file-wg_scope-md)

### DIFF
--- a/explainers/wpub-explainer.md
+++ b/explainers/wpub-explainer.md
@@ -7,7 +7,6 @@ Many types of publications consist of an ordered sequence of resources, which ma
 
 We also define serializations of this model in JSON-LD (and YAML?), with a vocabulary largely based on schema.org. Initial work will focus on audio books, but we plan to extend the model to encompass web publications, digital sequential art such as comics, manga, and bandes desinees, and  possibly even EPUB. 
 
-
 ## Goals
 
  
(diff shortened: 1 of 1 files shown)


iherman:
I like it. However, I do believe that some easy and quick editorial work should be done on the current documents before we contact the TAG. The name of the main document should be changed, the affordances should be parked somewhere, etc. I think these changes can be done quickly (@mattgarrish ?), we can then republish the document within a few days.

Also, depending on when the TAG review is scheduled, the remark on packaging may be updated. If indeed the decision will be to use OCF as a basis of this 'light-weight' packaging, then we should make it clear and also emphasize that this is the format already used by publishers, ie, the machinery to deploy it becomes easy and it is (will be?) well specified as a format.

mattgarrish:
> I think these changes can be done quickly

I got as far as making a branch to work on this yesterday... :)

Changing the name and stripping the affordances is certainly simple, but actually running through the specification and making sure everything makes sense to this new world view might take a little more time. I've got a couple of priorities off my plate now, though, so will see what I can accomplish this afternoon.

iherman:
I think that for a change/PR of such depth it would be worth leaving it open 1-2 days more and possibly notify the WG by email (as we will not have a meeting for 10 more days).

iherman:
@wareid, 

https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/394.html

is a URI for the updated WP draft done by @mattgarrish. It is not perfect re styling, but the content is there. So the mail could refer to this and the other PR in one blow.

iherman:
Oops, sorry @wareid , I forgot about the URI for _this_ PR without diff view:

https://github.com/w3c/wpub/blob/cc98b79356b6080c56421bb66eca792ad3baafda/explainers/wpub-explainer.md (a.k.a. https://tinyurl.com/y7pwnama)

iherman:
Per [mailing list cfc](https://lists.w3.org/Archives/Public/public-publ-wg/2019Jan/0009.html) merging and closing.

PR #397, opened 2019-01-30
Adding section to audio explainer on the basic outline of the format

dauwhe:
During the PBG call today, we were talking about getting feedback from audio publishers. Looking at the audio explainer with that audience in mind, I thought it would be helpful to add a very basic section on what the format would look like, complete with an illustration of a folder containing an audio book.


explainers/audio-explainer.md
```diff
@@ -34,6 +34,19 @@ The Audiobooks Task Force has identified the definition of an essential audioboo
 * A format that can be streamed, offlined, and downloaded
 * A format with clear metadata supported for user libraries (title, author, narrator, cover, duration, etc.)
 
+
+## How It Works
+
+The basic idea is simple: put all the audio files, a cover image, and some metadata in a folder, and zip the folder.
```
wareid: I would mention supplemental content here as well.

Issue #354, opened 2018-10-29
Proposal for handling localizable texts (writeup of the F2F discussions)


iherman:
This is an attempt for a write-up for the result of the [discussion on language and direction](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2018/2018-10-22-pwg.html#section5) of the F2F in Lyon on how to handle localizable texts in the Web Publication Manifest. (Note that this issue is really a JSON-LD one, hence the cc below to people outside of the WG.)

## Global setting of language and base directions

The [current draft](https://w3c.github.io/wpub/#manifest-default-language-and-dir) remains unchanged: we use the schema.org [`inLanguage`](https://schema.org/inLanguage) and our own `inDirection` term. The only change is that, eventually, the latter may become a schema.org term, and it should be removed, eventually, from the [WPM Context](https://www.w3.org/ns/wp-context)

## Item specific language (ie, Localizable text)

The mechanism that we agreed upon is as follows. There are three ways of setting a (localizable) literal. These literal are used for the title (ie, `name`) property, the names of creators, name of publishers, or (alt text) descriptions.

The three possibilities are, in an increasing level of complexity:

1. Simple string, ie,  
    `"name":"some text"`    
    where the value is a text in UTF8. 
2. String with explicit language setting, i.e.,  
   ```json
   "name": {
       "value": "some text",
       "language": "en"
   }
   ```  
   where `value` is a text in UTF8, and `language` may use any valid bcp47 tag.
3. Strings making use of the HTML control for language features, ie,  
   ```json
   "name" : {
       "value": "some text with <span>HTML syntax</span>",
       "datatype": "HTML"
   }
   ```
   where `value` is an HTML snippet. We will have to define a very minimal subset of HTML that fits the purpose of internationalization, and require that `value` MUST NOT go beyond that. The exact set of allowed HTML tags and attributes are still to be decided (hopefully in cooperation with other parties), but it will probably mean restricting to the usage of `span`, `ruby`, `rt`, `rb`, `bdi`, and `bdo` elements, and the `dir` and `lang` attributes.

---

Notes:

- Both in (1) and (2) Unicode's directional control characters may be used to control directionality, if necessary (see [appendix](https://w3c.github.io/wpub/#app-bidi-examples) for examples).
- Bcp tags may give more information than 'just' the language, and may also indicate, say, the script used. Example are `zh-Hant` and `zh-Hans` for Chinese written in traditional, resp. Simplified, characters. 
[...]

PR #407, opened 2019-03-06
Split the specification into parts

mattgarrish:
Opening the PR for general review and commenting, as I've gone as far as I'm likely to on my own in terms of splitting the document up.

There are now three parts to the specification:

- Part I defines the manifest in general terms for any "digital publication" (I had to introduce this term to avoid confusion, but maybe there's something better)
- Part II is the more familiar concrete implementation of web publications
- Part III will be where we define any requirements for extending

One place where I really noticed my attempts to generalize the manifest failing was with the table of contents and page lists, which depend entirely on there being a primary entry page. I did my best to allow flexibility for now, but as we've defined relationships for these I'm not sure what their actual property names would be. Will need revision in the future.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/407.html" title="Last updated on Mar 7, 2019, 7:07 PM UTC (eb5e46d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/407/dcd922c...eb5e46d.html" title="Last updated on Mar 7, 2019, 7:07 PM UTC (eb5e46d)">Diff</a>

### DIFF
--- a/common/css/common.css
+++ b/common/css/common.css
@@ -1,4 +1,18 @@
 
+section.part > h2 > span.secno {
+	display: none;
+} 
+
+a.suppress {
+	display: inline-block !important;
+	margin-left: -4rem !important;
+	margin-bottom: 0.5rem !important;
+}
+
+a.suppress > span.secno {
+	display: none;
+} 
+
 ul.ack > li,
 ul.flat > li {
     display: inline;
@@ -19,8 +33,7 @@ ul.flat {
 	padding-left: 0;
 }
 
-:not(div) > dl > dt,
-:not(div) > dl > dd {
+dt, dd {
 	padding-top: 0.5em;
 }
 
--- a/common/js/parts.js
+++ b/common/js/parts.js
@@ -1,0 +1,13 @@
+
+function denumber_parts() {
+	
+	var toc_entries = document.querySelectorAll('nav#toc > ol.toc > li.tocline');
+	
+	for (var i = 0; i < toc_entries.length; i++) {
+		toc_label = toc_entries[i].querySelector('a.tocxref');
+		
+		if (toc_label.textContent.match(/part i+:/i)) {
+			toc_label.classList.add('suppress');
+		}
+	}
+}
--- a/index.html
+++ b/index.html
@@ -7,15 +7,15 @@
 		<script src="common/js/biblio.js" class="remove"></script>
 		<script src="common/js/orcid.js" class="remove"></script>
 		<script src="common/js/wp.js" class="remove"></script>
+		<script src="common/js/parts.js" class="remove"></script>
 		<link href="common/css/common.css" rel="stylesheet" type="text/css" />
 		<script class="remove">
 		  // <![CDATA[
           var respecConfig = {
-			  postProcess: [show_orcid, create_wp],
+			  postProcess: [show_orcid, create_wp, denumber_parts],
 			  wg: "Publishing Working Group",
               specStatus: "ED",
               shortName: "wpub",
-              subtitle: "Manifest and Structure",
               previousPublishDate: "2019-01-30",
   			  previousMaturity: "WD",
               edDraftURI: "https://w3c.github.io/wpub/",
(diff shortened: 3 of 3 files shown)


HadrienGardeur:
> One place where I really noticed my attempts to generalize the manifest failing was with the table of contents and page lists, which depend entirely on there being a primary entry page.

We could simply reference the rel values that can identify them in `readingOrder` and `resources`, getting rid of the rest of it (this would be moved to a separate "Web Publications" document).

iherman:
General remark: I think if there is an agreement with the general direction, we should merge this ASAP. There may be technical issues, but it would be way better to handle those via bona fide issues.

In the meantime I will add two issues below which may or may not require further discussion.

---

Minor editorial thing:

- 2.2.: "(i.e., prior to be digital publication being processed by a user agent." Closing parenthesis missing, but I am not even sure the sentence is all right

iherman:
2.2.: says that each profile can define its own canonical manifest. I am not sure I like that. I would think that there _is_ a standard way to do that (like now) and that profiles have the possibility to _extend_ this when they provide new properties that require additional processing (which must be documented of course). This will require some editing on the canonicalization to describe extension points, but that is for later.

B.t.w., the current setup is also problematic with the WebIDL, because the WebIDL essentially represent the _canonical_ format and not the authored one.

iherman:
3.1 says 

> A Web Publication is discoverable in one of two ways: resources either include a link to the manifest (via an HTTP Link header or an HTML link element [html]), or the manifest can be loaded directly by a compatible user agent.

If we are talking about a _Web_ publication, then we did discuss and decide that one can discover one via the HTML entry page _only_. I think we decided _not_ to go into getting the manifest directly for all kinds of reasons (security, non-WP aware browsers, etc). I think that for Web Publications we must keep to that.
    
I realize that the package document allows a slightly different approach when things are packaged, but that is probably a separate, and it is not clear to me whether a separate package _is_ a module (I am not sure it is).

iherman:
Semi editorial issue: while we are making a major editing, it may be worth taking care of Richard's remark in https://github.com/w3c/wpub/issues/354#issuecomment-463716336: in 2.6.4.3.1 the description should say

> auto: indicates that the textual values are explicitly directionally set to the direction of the first character with a strong directionality, following the rules of the Unicode Bidirectional Algorithm [[!bidi]].

mattgarrish:
> it may be worth taking care of Richard's remark in [#354 (comment)](https://github.com/w3c/wpub/issues/354#issuecomment-463716336)

Sure, I didn't want to layer even more into this PR, but if that's all that needs to change I'll add.

mattgarrish:
> (i.e., prior to be digital publication being processed by a user agent.

Yes, this should be:

> (i.e., prior to the digital publication being processed by a user agent).

mattgarrish:
> If we are talking about a _Web_ publication, then we did discuss and decide that one can discover one via the HTML entry page _only_.

Right, I'm guilty of not rereading the intro to that part as closely as it was hoping it was just pre-existing content, but looks like a leftover from when we were still deciding about which options to support. I'll edit that out.

mattgarrish:
> 2.2.: says that each profile can define its own canonical manifest. ... I would think that there _is_ a standard way to do that (like now) and that profiles have the possibility to _extend_ this

I don't say that, I don't think. I said that the rules for _generating_ the canonical manifest are defined per format, and I'm not sure how we can avoid that.

How can you define common canonicalization rules without defining the content of a format that doesn't yet exist, since you have to be able to harvest info from it?

What we'd probably be looking at is defining is a generic publication format to go with the manifest.

iherman:
@mattgarrish, on https://github.com/w3c/wpub/pull/407#issuecomment-470674480: I think that even if we say "generating the canonical manifest are defined per format" is a bit loose and can lead to problems with interoperability.

Without having in mind all the details, here is what I thought could be done: what canonicalization does is to define mappings from shortcuts to full values. Thinks like:

- values that are strings and should become internationalized strings
- values that are single values and should become arrays
- values that are single values and should be put into a specific type of objects

etc.

What if we do the following:

- we categorize each relevant term according to the actions that have to be done
- the (generic) canonicalization algorithm, which currently has the term names baked in, would be re-written in terms of steps to be performed on certain categories
- there would be a possible extension point where format specific rules can also be added

With such a generic framework we would categorize the generic terms we have now and the canonicalization would then have the same effects as now; if a module adds new metadata values, they would have to add the relevant category, too.

WDYT?

On the practical side: I think we should flag it in the text and (if otherwise everybody is fine with it) merge this PR before getting into this for the next version.

mattgarrish:
If you mean generalizing the "generating" step, then we can probably make it work. I was thinking you wanted to move the entire lifecycle section into the manifest, and I'm not sure that's feasible given inevitable differences in how the manifest is obtained (e.g., finding from a web pub versus a package).

iherman:
@mattgarrish yes! My apologies, I wasn't clear...

mattgarrish:
Okay, then how about we merge this PR and then I can try some more changes? It would be nice to have a clean reset of the diff/preview before tackling more.

iherman:
@mattgarrish absolutely. This is what I proposed before: merge it and start from there!

iherman:
That being said: do we want to get an approval from our bosses? This is a significant change...

Cc: @TzviyaSiegman @wareid @GarthConboy

mattgarrish:
> That being said: do we want to get an approval from our bosses?

Ya, that's what I'm hoping for. I don't want to approve my own changes, so are we good to go?

wareid:
Looks good to me 👍

TzviyaSiegman:
Thanks @mattgarrish. This looks good.

iherman:
@mattgarrish I think we are at the point of: 'ship it'! :-)

We can then look at the canon. section separately.

PR #412, opened 2019-03-19
Added a text whereby using the embedded manifest is RECOMMENDED.

### LINKED ISSUE #327: Only allow embedded manifests
Many of the recent issues and discussions have been around the interplay between the primary entry page and the potentially external manifest. Much of the lifecycle and canonicalization content exists primarily to consolidate the various scenarios possible when externalizing the manifest.

Proposal: require that the manifest MUST be included in the primary entry page which is itself returned in response to a request for the publication's canonical address.

The result is that when you get a publication address, you get (in a single response) the means to understand and use the publication.

iherman:
This change is per https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-03-18-pwg#resolution2

Fix #327

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/412.html" title="Last updated on Mar 19, 2019, 4:03 PM UTC (324c4fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/412/46e24ce...324c4fc.html" title="Last updated on Mar 19, 2019, 4:03 PM UTC (324c4fc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1605,9 +1605,10 @@ dictionary CreatorInfo {
 									directionally set to left-to-right text;</li>
 								<li><code><dfn>rtl</dfn></code>: indicates that the textual values are explicitly
 									directionally set to right-to-left text;</li>
-								<li><code><dfn>auto</dfn></code> indicates that the textual values are explicitly directionally set
-									to the direction of the first character with a strong directionality, following the
-									rules of the Unicode Bidirectional Algorithm [[!bidi]].</li>
+								<li><code><dfn>auto</dfn></code> indicates that the textual values are explicitly
+									directionally set to the direction of the first character with a strong
+									directionality, following the rules of the Unicode Bidirectional Algorithm
+									[[!bidi]].</li>
 							</ul>
 
 							<p>When specified, these properties are also used as defaults for textual values in the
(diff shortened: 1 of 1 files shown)


mattgarrish:
Generally looks fine to me. I just made a couple of minor editorial tweaks to put the recommended practice first (emphasize what we want people to do) and to shorten the note.

iherman:
> Embedding is the preferred option as search engines will only process schema.org metadata in JSON-LD format when it is embedded in an HTML page.

May not be o.k. This makes a general statement on all search engines, which we cannot be sure is the case. Well, we know that this is what Google does (ie, does not), but on long term we cannot be sure and maybe, say, Yandex is more flexible.

What about a will->may, like:

> Embedding is the preferred option as search engines may only process schema.org metadata in JSON-LD format when it is embedded in an HTML page.

mattgarrish:
I actually took "may" out to avoid an rfc keyword in a note. :)

What about "typically" or "usually" or something along those lines rather than "may"?

iherman:
Is it a problem with 'may', as opposed to '_MAY_'? I do not think so... let alone the fact that this is a note, ie, informal anyway...

mattgarrish:
Well, yes, we should avoid keywords in informative content. Relying on uppercase to determine what is a keyword isn't terribly accessible.

It also adds confusion about whether we intended to make a normative statement, IMO, even if in this case it would be outside the scope of the spec.

In these cases, "might" can usually be dropped in instead with no impact on the meaning.

iherman:
So... can we use

> Embedding is the preferred option as search engines might only process schema.org metadata in JSON-LD format when it is embedded in an HTML page.

I made a new commit accordingly.

PR #419, opened 2019-03-27
Move manifest lifecycle to Part 1

mattgarrish:
This is essentially the second part of the process of abstracting the manifest. It takes the lifecycle section and makes it applicable to any module by moving it to the manifest section (Part 1) so it can be extended from there.

In particular, it:

- separates the process of obtaining the manifest, which will differ between web pubs and packaged formats
- creates the following categories of values with rules on how to process them:
  - literals - immutable strings (no processing)
  - implied and explicit objects - explanation on making localizable strings, "entities" (Person or Org) and links into objects is explained in the subsections
  - URLs - explains process for making relative in absolute
  - arrays - for properties that accept one or more of the above
- an additional "Value Category" is added to the definition tables - this serves to clarify which processing rule each property falls under, and also should help with explaining a single value or array is allowed.
- generalizes some of the canonical manifest generation steps and adds extension points
- defines some rules on how profiles should define profile-specific attributes

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/419.html" title="Last updated on Mar 27, 2019, 6:07 PM UTC (5613492)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/419/44750a8...5613492.html" title="Last updated on Mar 27, 2019, 6:07 PM UTC (5613492)">Diff</a>


iherman:
@mattgarrish forgot in the comment:

- defines some rules on how profiles should define profile-specific attributes

iherman:
@wareid can we merge? (I know @GarthConboy is away this week.) I would be good to do this before the Monday meeting that will give an overview of the latest changes...

wareid:
Looks good to me! Merge away!

Issue #420, opened 2019-04-01
Is duration required for audiobooks?


wareid:
As discussed in the call on April 1, is top-level duration of an audiobook required? 

Pros:
* Provides UAs with important metadata about the audiobook for processing
* Provides UAs with essential metadata to show user

Cons: 
* Duration metadata can potentially be inaccurate if left to content creator, causing issues for the UA

There might be other pros/cons, but I am leaning towards this being a requirement.

llemeurfr:
After internal discussion, we see the interest of having `duration` mandatory. It is not only a descriptive metadata used for display only, but also data used by the user agent to display a timeline for the audiobook (the duration of individual tracks is also really useful in this case, to get the relative size of each track inside the complete timeline).

BigBlueHat:
Based on our discussion today, what about this as a proposed option for closing this...

PROPOSAL: use https://schema.org/timeRequired on WP documents (of all kinds) to express expected "duration" of the publication, and continue to use https://schema.org/duration for a more technically accurate expression of length of a MediaObject or Audiobook type specifically.

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-04-08-pwg.html#section3).

- `RESOLVED:  Move the duration property to web publications`
<details><summary><i class="differentiate">View the transcript</i></summary>
duration in the context of WPUB<br/>
Wendy Reid: <a href="https://github.com/w3c/wpub/issues/307">https://github.com/w3c/wpub/issues/307</a><br/>
Wendy Reid: <a href="https://github.com/w3c/wpub/issues/420">https://github.com/w3c/wpub/issues/420</a><br/>
Wendy Reid: duration for entire content vs. duration at the resource level<br/>
Ivan Herman: duration currently defined in audiobook draft - nothing in there for duration is audiobook specific<br/>
… ability to add duration to a resource is generic - doesn’t have to be audio could be video or whatever else<br/>
… does it make sense to have a duration for the book as a whole? for audio books, it is a requirement, but not for a general web publication. but this may not be restricted to audio books. Maybe there could be a video book. The concept is generic, but it may be a requirement for audiobooks and not for general web publications.<br/>
Avneesh Singh: generic, but in audio profile should be compulsory. But what is duration of whole book? Does that include branches that are not included in the reading order?<br/>
[...]

Issue #307, opened 2018-08-16
Duration of an audiobook


HadrienGardeur:
For an audiobook, we need to ideally express:

- the total duration of the publication
- the duration of individual audio resources

Schema.org uses https://schema.org/duration for this purpose but there are a few points worth listing:

- ISO 8601 for a duration is not always the friendliest format
- media fragments ([used in a TOC for example](https://github.com/w3c/wpub/blob/master/experiments/audiobook/toc.html)) are based on seconds instead
- it would require the use of `"@type": "Audiobook"` as well, since `duration` is not defined on `CreativeWork` or `Book`
- it's also undefined right now on `PublicationLink` or the equivalent in schema.org that's currently a draft (I can't remember the exact term)

For the duration of individual resources listed in the reading order, the same issue raised in https://github.com/w3c/wpub/issues/229#issuecomment-413520551 applies as well.

danielweck:
Thanks for pointing this out.

Indeed, I am discovering the "duration" ISO_8601 syntax:
https://en.wikipedia.org/wiki/ISO_8601#Durations

I am more familiar with Media Fragments:
https://www.w3.org/TR/media-frags/#naming-time

...and of course SMIL (well, the EPUB3 Media Overlays simplified model):
http://www.idpf.org/epub/31/spec/epub-mediaoverlays.html#app-clock-examples
https://www.w3.org/TR/SMIL3/smil-timing.html#Timing-ClockValueSyntax

llemeurfr:
You're right @HadrienGardeur, the ISO 8601 syntax for a duration is un-intuitive (but easy when you look closely) and I didn't see it used in the read world (e.g. real life XML schemas). 

We must consider that this duration property is essentially used for display purposes (in the UA), therefore the second is a sufficient precision and an audiobook duration will never be much more than 24 hours. A duration is minutes or seconds would therefore be ok.

I'm sligthly in favor of reusing the SMIL timecount syntax (e.g. `45min`, or `30s`). Easy to write, to read and to parse. I understand this is a fork from schema.org. Nb: using the ISO 8601 syntax we would have values like `T45M` or `T30S`.

Apart from the global duration of the audiobook (which will appear along with the description of the work), my feeling is that the duration of individual "chapters" should be expressed at the level of the ToC rather than the level of individual (physical) resources.

iherman:
[...]

PR #421, opened 2019-04-09
Added duration to the LinkedResource object type

iherman:
This is the implementation of the relevant [resolution](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-04-08-pwg#resolution2) on the meeting of the 4th of April.

This PR _only adds the resource level duration_. The book level term is not yet decided, see #420, more exactly the [proposed solution](https://github.com/w3c/wpub/issues/420#issuecomment-480915090). If that is also decided, I am happy to add a book level value, too.


HadrienGardeur:
@iherman here are the duration of the tracks for the Flatland example in another syntax: https://github.com/readium/webpub-manifest/blob/master/examples/Flatland/manifest.json

danielweck:
Yes, the Schema.org definition of `duration` ( https://schema.org/duration ) / `Duration` ( https://schema.org/Duration ) references the "duration" syntax (i.e. not date, time, interval, etc.) of ISO 8601 ( https://en.wikipedia.org/wiki/ISO_8601#Durations ).

Hadrien gave a link that shows the ReadiumWebPubManifest `duration` JSON property, which is a decimal `number` of seconds (i.e. not a `string` expression). This is a tiny subset of the full "temporal dimension" syntax allowed in Media Fragments URIs ( https://www.w3.org/TR/media-frags/#naming-time ).

If I remember correctly, a reference to RFC 7826 ( https://tools.ietf.org/html/rfc7826 ); as mentioned by Ivan; was introduced because Media Fragments references RFC 2326 ( https://tools.ietf.org/html/rfc2326 ) which RFC 7826 explicitly supersedes.

All being said, we've got to pick one *or* the other :)
* ISO 8601 "duration" format (because Schema.org)
* RFC 7826 Normal Play Time (because Media Fragments URI)

iherman:
Thanks @danielweck... 

> All being said, we've got to pick one _**or**_ the other :)
>
> * ISO 8601 "duration" format (because Schema.org)
> * RFC 7826 Normal Play Time (because Media Fragments URI)

I agree we must do this. My vote would be to go with the first approach. We have tried to align, as much as we can, with schema.org, we even refer to https://schema.org/duration, which tells me that we do not really have a choice there. I do realize the syntax is more clumsy, but…

iherman:
> @mattgarrish I am not sure this PR should be merged before or after the draft is republished as /TR. Per our resolution this could/should be done at any time...

Publication is done now, i.e., if the other issues are solved, and the PR is approved, we can also merge.

llemeurfr:
@iherman the choice of a number of seconds (decimal figures allowed) has been a pragmatic choice made by the Readium community: everybody understands such format, and media durations are never in the range of years, months or days (this is what the ISO 8601 format is targeting). 

We had this discussion before in https://github.com/w3c/wpub/issues/307 and the conclusion was "after discussion with schema, we will use NPT (Media Fragments) until they confirm or inform us otherwise".

iherman:
@llemeurfr, if that is the WG final decision, than that is what we will do. However, in this case, we should _not_ refer to schema.org, nor should we use the same term to avoid problems with JSON-LD processors (which default unqualified terms to schema.org).

iherman:
> Note that we should be more precise and state we use npt-sec (not npt-hhmmss)

why? 

I am a bit wary of the direction. If we use NPT from an RFC, then we should use the NPT as a whole. Just cherry picking part of a standard is a source for problems later imho.

llemeurfr:
@iherman 

>  If we use NPT from an RFC, then we should use the NPT as a whole. 

After all why not ... the syntax is so that the parser can easily make the difference, and for expression the expected duration of a textual ebook, hours + minutes make sense. 

npt-sec      =   1*DIGIT [ "." *DIGIT ]
   npt-hhmmss   =   npt-hh ":" npt-mm ":" npt-ss [ "." *DIGIT ]
   npt-hh       =   1*DIGIT     ; any positive number
   npt-mm       =   1*2DIGIT    ; 0-59
   npt-ss       =   1*2DIGIT    ; 0-59

exemples: 
using npt-sec: 123.45
using npt-hhms: 12:05:35.3

llemeurfr:
Also, the conclusion of #307 was that we would request schema.org to allow this NPT notation as an alternate syntax on Duration, which would then allow us to use the schema.org property (either `duration` of `timeRequired`). But this didn't happen so far.

iherman:
> Also, the conclusion of #307 was that we would request schema.org to allow this NPT notation as an alternate syntax on Duration, which would then allow us to use the schema.org property (either duration of timeRequired). But this didn't happen so far.

I.e., it must be done on the schema.org mailing list. But there has to be a good argument for it, e.g., what is wrong with the ISO formalism, evidence that it does not work, etc. It would make the case stronger if we had evidence of widespread usage of NPT on the Web as opposed to the ISO version. My gut feeling is that it will be a difficult call to convince schema.org to allow for a dual syntax.

danielweck:
> It would make the case stronger if we had evidence of widespread usage of NPT on the Web as opposed to the ISO version. 

Well, W3C Media Fragments URI is a web specification that references the RFC, not the ISO duration format, so even without usage metrics the Schema.org choice is arguably less "Webby" :)

danielweck:
...and the W3C SMIL standard also utilises NPT notation for clock values.

danielweck:
For the sake of completion / exactitude, here is the spec. for the SMIL clock values syntax:

https://www.w3.org/TR/SMIL3/smil-timing.html#q22

...and note that ISO 8601 is actually also utilized in SMIL, albeit for wallclock values which are much less common (and not used in EPUB3 Media Overlays, for example):

https://www.w3.org/TR/SMIL3/smil-timing.html#Timing-WallclockSyncValueSyntax

iherman:
(Trying to move forward...) I would propose the following steps of actions

1. The current draft is modified, _removing_ the reference to the ISO format and leaving NPT as the only reference and then, unless there are other issues, we merge this PR to  make it in the latest draft
2. We (maybe temporarily?) use a different term instead of `duration` and `timeRequired` (I am not sure what they should be, maybe the time-based media world has some alternate terminology.) Ideas for these terms are welcome…
3. We send a (well argumented…) request to the public-schemaorg@w3.org list asking to _extend_ the accepted values for `duration` and `timeRequired` to include NPT. If that is accepted, we can name our terms (back) to use the schema.org ones.

@llemeurfr @danielweck @wareid @GarthConboy @TzviyaSiegman @mattgarrish : WDYT?

danielweck:
> WDYT?

SGTM

iherman:
But... before I merge, I need an alternative name to `duration`...

laudrain:
some proposal : timeLength, contentLength, consumptionTime, escapeTime, TimeToEnd

danielweck:
`dur`

mattgarrish:
@danielweck beat me to it, but `dur` is used in the NCX and SMIL. Good enough for now.

iherman:
`dur` is fine with me for now (unless we get schema.org's blessing for using NPT).

Shall we reserve `contentLength` to replace `timeRequired` if we otherwise agree on #420?

iherman:
@HadrienGardeur thanks, I will use those figures in the example.

mattgarrish:
Using `timeRequired` for total duration bothers me, to be honest. It's an LRMI term for describing how long it will typically take a student to work through a learning resource. It just feels like a misappropriation.

But `totalDur` for symmetry would be fine with me.

llemeurfr:
@iherman, possible alternatives =`readingDuration`, `readingTime`,  `approximateDuration`.

wareid:
> Using `timeRequired` for total duration bothers me, to be honest. It's an LRMI term for describing how long it will typically take a student to work through a learning resource. It just feels like a misappropriation.
> 
> But `totalDur` for symmetry would be fine with me.

I feel like you hit my discomfort with the term on the head, I just wasn't able to articulate it. 'totalDur' sounds much better to me.

danielweck:
@iherman 
> Shall we reserve contentLength to replace timeRequired ...

High risk of conflation with HTTP `Content-Length`:
https://tools.ietf.org/html/rfc7230#section-3.3.2

danielweck:
@llemeurfr 
>  possible alternatives = readingDuration, readingTime

I prefer an abstract term like `totalDur` vs. experience-specific terms like `readingDur` and `listeningDur`.

mattgarrish:
Another alternative might be to use dcterms:extent as the interim solution so we don't have to completely mint something new.

llemeurfr:
@danielweck ok for `totalDuration` or `globalDuration` then; we should no make property names difficult to understand just to save 5 characters.

danielweck:
Of course I prefer "duration" to "dur" :)
(which in French means "hard", btw)

I was just echoing Matt's concern about consistency:
> But totalDur for symmetry would be fine with me.

marisademeglio:
Do we need two different words for duration of a single resource vs duration of an entire publication? Not to be that person who says "well this is what we did in EPUB", but this is what we did in EPUB ;) 

https://w3c.github.io/publ-epub-revision/epub32/spec/epub-mediaoverlays.html#example-7

Can we use only `duration` and if it's on a resource, it applies to that resource; and if it's at the root-level (along with other publication-wide properties), it applies to the entire publication?

iherman:
@marisademeglio 

> Do we need two different words for duration of a single resource vs duration of an entire publication? 

This is not yet a decision, it has to be discussed separately on #420. It only came up here as part of the naming of the term.

iherman:
.I have made a new commit. The automatic preview seems to be down, the human readable version of the version is here:

https://raw.githack.com/w3c/wpub/adding-duration/index.html

What I did, in view of the previous comments:

- I used, temporarily, the term `totalDuration`, per https://github.com/w3c/wpub/pull/421#issuecomment-481844907. Obviously, the term can be changed at any time, but having a consistent document may be good at this point.
- I have also added (see []()) the `globalDuration` as a separate, descriptive terms. The same note as above applies.
- For both terms:
    - I have removed the reference to the ISO duration value
    - Added a note whereby we will be talking to schema.org whether it is possible to extend the accepted values to their terms to include NTP, in which case we will fall back on their terms.
- I have also added a reference to the open issue #420 into the document so that we would not lose track of it.
- I have also taken over the duration values provided by Hadrien for the audiobook examples.
- The post-processing of the manifest includes two more possible warnings, namely if the value of `globalDuration` or `totalDuration` is not a valid NPT. (The relevant diagram has also been updated.)

@mattgarrish @TzviyaSiegman @wareid @GarthConboy @llemeurfr @danielweck is it o.k. to merge it?

llemeurfr:
@iherman, the *duration values should be JSON numbers (i.e.decimal numbers = "floats"), not strings IMO.

iherman:
@llemeurfr I was wondering about that, but, e.g., "12:05:35.3" is a valid NPT value, and that is not a number...

What we could do is to say that the user _MAY_ use a number instead of a string (and, e.g., the canonicalization would turn it into a string). This can easily be done, but I am not sure it is good idea to complicate things (implementation may accept a number under the motto "be liberal what you accept"...)

llemeurfr:
@iherman, you're right, I was forgetting the npt-hhmmss variant. So string be it ...

geoffjukes:
@iherman Late to the party, and this is a long thread.

What is the objection to simply stating the duration in seconds? why would we need or want to have a formatted string?

iherman:
@geoffjukes 

At the moment, the value is defined to be NPT as defined by rfc7826. _If_ we do refer to a standard in this area, then we must take it fully, i.e., we should allow for "12:05:35.3" as well, which is a string.

From an editor's perspective, I can see two different possibilities

1. We stick to NPT, but we allow for the value to be _either_ a number _or_ a string, where the number _**N**_ means the same as if it was _**"N"**_ (which is possible and valid NPT and which is a fairly common JSON pattern).

    (Editorially, what I presume this may mean is that the _authored_ manifest would allow for a number, whereas the canonicalization would turn this into a string for the canonical manifest.)
2. We _remove_ any reference to NPT, or any kind of standards altogether, and we _define_ the value to be simply the (possibly fractional) number of seconds.

I am not an expert in audio books; my gut feeling is that allowing the (1) may be more comfortable for authors, hence I would prefer to go down that way.

---

If we are able to harmonize the duration value with schema.org, it looks like schema.org processors accept a number where they expect a string with a number (e.g., for copyrightYear), so (1) is not a problem in this sense. (I made some test on the [Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool).) 

Cc: @mattgarrish

iherman:
B.t.w. the diff file can be extracted here: https://tinyurl.com/y3suyhcm

(It takes a while, though...)

geoffjukes:
@iherman `duration` is more commonly a double, rather than a string. If string formatting is required/desired, it is handled by the UI.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/duration

```
The HTMLMediaElement.duration property gives the length of the media in seconds, or zero if no media data is available
```

https://docs.microsoft.com/en-us/windows/desktop/wmp/media-duration

```
Media.duration
The duration property retrieves the duration of the current media item in seconds.
Possible Values
This property is a read-only Number ( double).
```

https://developer.android.com/reference/java/time/Duration

```
A time-based amount of time, such as '34.5 seconds'.
```

We use the word `length` rather than `duration` (because the creation tools we use do).

geoffjukes:
@iherman there is some overlap between this ticket and #420, and I apologize for contributing to that.

I think there is a distinction between `duration` and `timeExpected` that is important to consider.

It may be more appropriate to format `timeExpected` as an NPT string (as it's subjective time and user-facing) and a media file `duration` as a value in seconds (as a `double`)

danielweck:
I think it is a false equivalency to compare (1) types used in APIs to read and/or write values, with (2) syntaxes used in authoring and interchange formats. They do not serve the same purpose.

Developers of processing / playback engines for timed media (i.e. audio/video, SVG animations, SMIL, etc.) are indeed very likely to normalize clock values into internal, implementation-specific data types that lend themselves to efficient calculations (such as: number of seconds with decimal point, subject to the constraints of floating point arithmetics).
However, `00:02:30.500` (NPT syntax) is arguably more expressive than `150.5` (seconds) for the purpose of information exchange and publishing. That is not to say that the latter should be forbidden, but rather that the former should be allowed.

There is prior art at W3C (e.g. SMIL, Media Fragments URI) which borrows from other RFC and ISO specifications. I think WebPub should support such well-established, expressive syntactical constructs based on strings of characters, whilst also allowing JSON numbers (floating point) to directly represent units of seconds. The net advantage of the latter is of course performance (no parsing / serialization costs), and; if I understand correctly; the normative definition of "canonical" WebPub can choose this.

References:

SMIL:
https://www.w3.org/TR/SMIL3/smil-timing.html#q22
https://www.w3.org/TR/SMIL3/smil-timing.html#Timing-WallclockSyncValueSyntax

Media Fragment URIs:
https://www.w3.org/TR/media-frags/#naming-time

And just for fun, here are HTTP links to a 3-minutes-long MP3 audio file, starting at 2 minutes 30 seconds and 500 milliseconds, using equivalent Media Fragment URI syntaxes:

`00:02:30.500`
http://narwakk.free.fr/musiques/Bob/Bob%20Marley%20-%20Roots,%20Rock,%20Reggae.mp3#t=00:02:30.500

`150.5`
http://narwakk.free.fr/musiques/Bob/Bob%20Marley%20-%20Roots,%20Rock,%20Reggae.mp3#t=150.5

(works in all modern web browsers)

geoffjukes:
@danielweck I think we have different interpretations of the use and purpose of this `duration` property, and I think this may be one of 'location'.

My understanding is this:

* A LinkedResource object represents a single discreet item with a variety of properties (duration being one, if appropriate).

* TOC/readingOrder (whatever it is now) may re-use linked resources, and should be capable of specifying a Media Fragment style duration (or whatever is agreed).

geoffjukes:
OK. I am *finally* on the same page here.

As commented on https://github.com/w3c/wpub/issues/420#issuecomment-482859276 the ISO 8601 "duration" format is the most appealing to me, as it allows for seconds-based expressions such as `PT130.5S`. I have confirmed that all the tools we use in production can create and parse it.

My primary concern/objection came from the misconception that we would need to format seconds into `HH:MM:SS.ssss` strings, which is not the same as `'PT{}S'.format(seconds)` 😄 

I think the media fragments discussion is more appropriate to a ToC/readingOrder discussion,

iherman:
@geoffjukes 

`"SS.ssss"` (i.e., a float number added as a string) is a perfectly fine value for NPT, afaik. As I said in https://github.com/w3c/wpub/pull/421#issuecomment-482470225 it would be perfectly possible to define the manifest in a way that the user/author could use a (float) number in term of a JSON datatype _or_ an NPT formatted string that could also include the float number as a string. Put it another way, it is not _required_ to use "HH:MM:SS.ssss" strings to express the length in seconds, it is just a convenience possibility.

I am not an expert, i.e., I will follow the choice between NPT and ISO as agreed by the Working Group.

llemeurfr:
@geoffjukes let's imagine that you have to explain to the whole publishing industry, including audiobook studios, that media duration (a track in audio speak) is not expressed as 130.5 (in sc) but as "PT130.5S" and that is normal because ... ISO. How many eyebrows will you face?
The ISO 8601 is great because it is very flexible (date, time, duration, time intervals, recurring durations, you can express many things) but for a property named duration it does not follow the KISS principle IMHO.

geoffjukes:
@llemeurfr from https://www.ietf.org/rfc/rfc2326.txt:

> 3.6 Normal Play Time
>
>   Normal play time (NPT) indicates the stream absolute position
>   relative to the beginning of the presentation

It doesn't read like it is intended for how you are proposing to use it. I'll go with what the majority decides (not I know I can comfortable express seconds as a decimal in both NPT and ISO 8601) but my preference would be ISO, and conformance with schema.org

iherman:
This issue was discussed in [a meeting](https://www.w3.org/publishing/groups/publ-wg/Meetings/Minutes/2019/2019-04-15-pwg#section2-1).

- `RESOLVED:  The time length property (unnamed) will only use a float consisting of the number of seconds of the resource.`
- `RESOLVED:  duration is a descriptive metadata for WP, whose value is the ISO format (as used in schema.org). It is optional.`
- `RESOLVED:  Schema.org duration value is _recommended_ metadata for the audiobooks profile`
<details><summary><i class="differentiate">View the transcript</i></summary>
duration<br/>
Wendy Reid: Issue: <a href="https://github.com/w3c/wpub/issues/420">https://github.com/w3c/wpub/issues/420</a><br/>
Wendy Reid: Open Pull Request: <a href="https://github.com/w3c/wpub/pull/421">https://github.com/w3c/wpub/pull/421</a><br/>
Wendy Reid: should duration be required?<br/>
… duration will become part of the core spec<br/>
Ivan Herman: there are 2 issues<br/>
… one is, if we take duration for one resource, the question that came up was what is the format for the value?<br/>
… one possibility would be to use the ISO 8601 Value, which is used by schema<br/>
… or we could use RFC 7826<br/>
… which is used in the media world<br/>
… the majority seem to favor the RFC value, as it’s more readable than ISO<br/>
… we had an issue #307 a while ago where that was decided, but it wasn’t clean in the doc<br/>
… we can reinforce this decision<br/>
… this is one part of it<br/>
Wendy Reid: I’ve tried to talk to danbri about this, no response yet<br/>
… the RFC value fits better with what we want to do, especially if we also want to reference media fragments<br/>
… can we merge the PR and close this issue?<br/>
Ivan Herman: do we want to make a new resolution, or the already decided one?<br/>
Wendy Reid: let’s stick with the NPT?<br/>
Geoff Jukes: I’m confused about the intent<br/>
… specifying the duration of the resource<br/>
… it’s the file, effectively<br/>
… that duration is only specified in seconds<br/>
… never anything else<br/>
… my concern is that putting in media fragments at the resource level doesn’t make sense<br/>
… if the intent is to conform to schema.org, and we should just use ISO<br/>
… so why NPT instead of using a double?<br/>
… I don’t know why it’s a consideration<br/>
Wendy Reid: media fragments will be a thing, although maybe more in TOC etc than in resources<br/>
Geoff Jukes: that’s not describing a resource, but metadata<br/>
Wendy Reid: we don’t want two different formats for these things<br/>
Geoff Jukes: I disagree<br/>
… and I’m having trouble with these very long discussions<br/>
… I think of a media fragment as a different thing than a resource<br/>
Deborah Kaplan: geoffjukes: +1 for calling out our confusing conversations as confusing. Thanks.<br/>
Ivan Herman: the NPT format is defined in a way that it can have only a number, which is seconds<br/>
… the author may choose to use raw seconds<br/>
Tzviya Siegman: i missed last week. is it possible to summarize the discussion?<br/>
Ivan Herman: in a way we jumped ahead<br/>
… we have make a choice between ISO and RFC<br/>
… then during the discussion a third option became possible, just taking the number of seconds<br/>
… those are the three options<br/>
… so the question is which of the three?<br/>
Geoff Jukes: in addition to that, what is the desire to conform to schema? Is that a design principle?<br/>
Ivan Herman: we want the contents of the manifest to be accessible to the knowledge graph<br/>
… it’s mostly important for bibliographic metadata<br/>
Geoff Jukes: the desire to conform to schema is high, so we can obtain cross-vendor parsing capability<br/>
… is that correct?<br/>
Ivan Herman: yes<br/>
Laurent Le Meur: here we are speaking on duration of resource, not duration of audiobook. It’s not a property of a book.<br/>
… so it’s not tied to a need to express audiobook metadata for schema<br/>
… so we could use seconds, with a name other than duration (like runtime or length)<br/>
… and the audiobook industry would be happy with that<br/>
Geoff Jukes: I’d be happy with a new thing called length or whatever, that’s just a double<br/>
… it’s what we already do<br/>
Ivan Herman: to be clear I am just a messenger<br/>
… whatever the group decides is fine<br/>
Tim Cole: the decision could be made, that in this community we would use duration but constrain the value of seconds<br/>
… i think this is OK<br/>
… it could be enforced via a context document<br/>
… we could also define our own property, and connect it to duration<br/>
… there are ways to express constrained versions of other properties<br/>
Ivan Herman: I don’t think that works<br/>
… schema uses the ISO format, and it doesn’t allow a simple number<br/>
… a number can be a subset of RFC, but not of ISO<br/>
Wendy Reid: the reason we were leaning on RFC it has only two ways to express time, including only seconds<br/>
Ivan Herman: I would propose to move on<br/>
… we define that property to have a value being a float consisting of the number of seconds<br/>
… with a new term like length<br/>
Proposed resolution: The time length property (unnamed) will only use a float consisting of the number of seconds of the resource. (Wendy Reid)<br/>
Ivan Herman: +1<br/>
Laurent Le Meur: +1<br/>
Dave Cramer: +1<br/>
Franco Alvarado: +1<br/>
Tim Cole: +1<br/>
Geoff Jukes: +1<br/>
Joshua Pyle: +1<br/>
Marisa DeMeglio: 0<br/>
Avneesh Singh: +-0<br/>
Avneesh Singh: no strong opinion :)<br/>
… waiting for feedback from media sync people<br/>
Wendy Reid: does this impact sync media<br/>
Marisa DeMeglio: I don’t think so… this is just properties of resources<br/>
Ben Schroeter: +1<br/>
Brady Duga: Abstain (don’t plan to use the value)<br/>
Marisa DeMeglio: this issue doesn’t need to get more complicated<br/>
Resolution #2: The time length property (unnamed) will only use a float consisting of the number of seconds of the resource.<br/>
Ivan Herman: the other issue that came up is more controversial<br/>
… there may be a notion of duration of the whole audiobook<br/>
… it turned out that having that as book-level metadata is something that implementors may ignore<br/>
… they may deduce that from the individual resources<br/>
… but it may be helpful as a hint to the user, as a value in the catalog etc<br/>
… what I did, mostly to generate discussion, was to<br/>
… put a global property there, with the same format<br/>
… primarily defined for a user interface<br/>
… do we need this, or should we remove it from the PR doc?<br/>
Laurent Le Meur: I would say that the audiobook schema.org object supports the duration with ISO 8601<br/>
… it’s there and it’s optional, and it is what we want<br/>
… it’s descriptive metadata<br/>
… we could just adopt this and move on<br/>
Ivan Herman: +1 to laurent<br/>
Geoff Jukes: +1 to laurent<br/>
Wendy Reid: simple descriptive metadata<br/>
Proposed resolution: Schema.org’s Duration will be a required metadata descriptor for audiobooks (Wendy Reid)<br/>
Ivan Herman: +1<br/>
Ben Schroeter: +1<br/>
Laurent Le Meur: I thought the idea was to keep it optional, as in schema.org<br/>
Proposed resolution: Schema.org’s Duration will be a recommended metadata descriptor for audiobooks (Wendy Reid)<br/>
Laurent Le Meur: and it’s a ‘duration’ property (of type ‘Duration’)<br/>
Ivan Herman: +1<br/>
Laurent Le Meur: +1<br/>
Marisa DeMeglio: is this a different property?<br/>
Ivan Herman: yes<br/>
Marisa DeMeglio: -1<br/>
Deborah Kaplan: +1<br/>
Tim Cole: +1<br/>
Joshua Pyle: +1<br/>
Wendy Reid: this is schema.org duration descriptor for audio book, the length of the entire work, the sum of all the parts<br/>
Laurent Le Meur: see <a href="https://schema.org/Audiobook">https://schema.org/Audiobook</a><br/>
Laurent Le Meur: and <a href="https://schema.org/duration">https://schema.org/duration</a><br/>
Geoff Jukes: it’s not the same concept<br/>
… it might be the sum of all resources, but it might be different, for example if there’s non-book audio resources<br/>
Ivan Herman: +1 to geoffjukes<br/>
Geoff Jukes: +1<br/>
Geoff Jukes: so it’s ok to have a different name and format, and it’s good for it to be in schema.org so it’s universally digestable<br/>
Wendy Reid: it would be called duration, it would be the total length of the book, provided by the publisher<br/>
Deborah Kaplan: are we voting on making this required?<br/>
Ivan Herman: there is a mess-up<br/>
… there are 2 things here<br/>
… one is, what is the global descriptive metadata, and what value it takes<br/>
… and the only resolution we are proposing is to use duration with ISO as in schema.org<br/>
… and then there’s the question of whether this metadata item is required<br/>
Geoff Jukes: I would happy for it to be required<br/>
… we have to send it to our publishers/distributors<br/>
Wendy Reid: when I said required I meant for the audiobook profile<br/>
Laurent Le Meur: q for geoffjukes. Why is it required?<br/>
Geoff Jukes: when we send ONIX we include runtime<br/>
… and they like to cross-reference to check they make sure they got the right book<br/>
… if it’s not required we’ll supply it anyway<br/>
Tzviya Siegman: +1 to limited metadata!<br/>
Dave Cramer: The web platform requires very little metadata, we should require the important things (title, author), this does not seem like required metadata<br/>
… I suggest we make it optional<br/>
Ivan Herman: +1 to dauwhe<br/>
Laurent Le Meur: +1 to dauwhe<br/>
Wendy Reid: for an audiobook it’s almost as important as title<br/>
… for a user to understand what they’re getting into<br/>
… to find out if it’s abridged or unabridged<br/>
… or if my phone will keep it<br/>
… I think it should be required<br/>
Ivan Herman: there is no requirement to provide metadata for the number of book pages<br/>
… but the same argument applies, ish<br/>
… I agree it is recommended<br/>
… but “must” is too far<br/>
Tzviya Siegman: I hate to prolong this discussion<br/>
… when we were deciding on EPUB metadata, lots of people said that title should be required<br/>
… but then you get into lots of nuance with what titles means, but most systems don’t pay attention<br/>
… we should look into how systems work with information about length<br/>
… and how this will play out in the real world<br/>
… maybe the implementors can tell us more about this information is used<br/>
Dave Cramer: It strikes me as many of the arguments for the utility of the information is about file size not chronological duration, this information can be useful, but requiring them is not traditionally how the web works<br/>
… we run into issues of validation<br/>
… are we then going to get to a point where validators takes the values and compares them<br/>
… requiring this is complicated<br/>
Laurent Le Meur: I agree on principle we shouldn’t require descriptive metadata<br/>
… and we should keep properties required for user agent functioning or content identification<br/>
… so we should recommend this, underlining all the advantages of using this<br/>
Brady Duga: when considering required metadata, we should ask if it’s impossible to create a book without this metadata.<br/>
… if it’s not impossible, we shouldn’t require it<br/>
Wendy Reid: I’m OK with recommended, even though y’all are completely wrong :)<br/>
Bill Kasdorf: vendors can still require it<br/>
Ivan Herman: we have 2 resolutions to take<br/>
… we never closed the previous resolution<br/>
Proposed resolution: duration is a descriptive metadata for WP, whose value is the ISO format (as used in schema.org). It is optional. (Ivan Herman)<br/>
Ivan Herman: +1<br/>
Tim Cole: +1<br/>
Laurent Le Meur: +1<br/>
Bill Kasdorf: +1<br/>
Deborah Kaplan: +1<br/>
Geoff Jukes: +1<br/>
Ben Schroeter: +1<br/>
Dave Cramer: +1<br/>
Brady Duga: +1<br/>
Garth Conboy: +1<br/>
George Kerscher: +1<br/>
Resolution #3: duration is a descriptive metadata for WP, whose value is the ISO format (as used in schema.org). It is optional.<br/>
Joshua Pyle: +1<br/>
Proposed resolution: Schema.org duration value is recommended metadata for the audiobooks profile (Wendy Reid)<br/>
Ivan Herman: +1<br/>
Laurent Le Meur: +1<br/>
Marisa DeMeglio: +1<br/>
Tim Cole: +1<br/>
Bill Kasdorf: +1<br/>
Ben Schroeter: +1<br/>
Deborah Kaplan: +1<br/>
Resolution #4: Schema.org duration value is _recommended_ metadata for the audiobooks profile<br/>
Wendy Reid: can we move on and never speak of this again?<br/>
Ivan Herman: no<br/>
… I will make the edits according to the resolutions, is it OK to then merge?<br/>
Wendy Reid: +1<br/>
Ivan Herman: and then the PR and the issue can be closed then?<br/>
<b>everyone:</b> YES<br/>
</details>

iherman:
@mattgarrish @wareid @TzviyaSiegman @GarthConboy 

I have made the changes we discussed yesterday on the call.

- Added a new value category for 'Number"
- Added the (optional) 'length' attribute to LinkedResources with Number
- Added the (optional) 'duration' descriptive metadata value with a value as ISO's Duration (I have added a note referring to the wikipedia page for duration for those mortals who cannot pay for the ISO document:-)
- Added, to the post-processing steps, a check for the duration and for length (compatibility to ISO for the former, valid number for the latter)
- Updated the audiobook examples to use the new terms
- Updated the diagram on the post-processing steps

If you guys agree, we should merge asap; there are already merge conflicts because, in the meantime, @mattgarrish also made some changes. 

(@mattgarrish, can you take care of those conflicts? I am not sure I understand, the conflict is on a WebIDL; I tried to take over the latest version to this document, but it still sees a conflict...)

---

- new version is accessible at: https://raw.githack.com/w3c/wpub/adding-duration/index.html
- diff is accessible at: https://tinyurl.com/y3suyhcm

mattgarrish:
Any particular reason for changing the URLs in the quick reference? I thought we were leaving those until the naming was resolved, or did I miss somewhere that it has been?

More why I ask is that we seem to be falling into the same trap we did with the epub structure vocab with inconsistent naming. The schema.org properties use camelcased names, but we also have a mix of hyphenated names (accessibility-report, privacy-policy) and lowercase (pagelist). Would be good to standardize on camelcasing.

iherman:
1. I realized that we do not put urls consistently in the quick reference; why would we put ours and not for schema? For the authors the difference is insignificant, the mapping to urls happens in the json-ld contexts for both. Isn't this stuff more readable
2. Agree on the camelcase thing. Which are the ones we will have to change?

mattgarrish:
> why would we put ours and not for schema

Because these are relationships that aren't yet standardized, not properties, so they were defined as URLs. I believe they all have notes in their respective definitions that we are looking to have them integrated in the appropriate vocabulary so the URLs don't have to be used (i.e., into IANA).

If you have to use the URL in the `rel` attribute, shouldn't we be consistent in pointing out that, at least for now, a URL is expected?

The bigger question might be why do we intermingle properties and relationships as though they're the same? It might be something to rethink.

Issue #426, opened 2019-04-16
Informative and structural "properties" are relationships


mattgarrish:
To separate this discussion away from #421, I find a bit awkward that we lump the resource relationships into the properties. These should really be a separate section, for example:

```
2.7 Resource Relationships

2.7.1 Informative Relationships
2.7.1.1 Accessibility Report
2.7.1.2 Privacy Policy

2.7.2 Structural Relationships
2.7.2.1 Cover
2.7.2.2 Page List
2.7.2.3 Table of Contents
```

The extensibility of relationships is also different from properties, since we expect these to be standardized in IANA.

mattgarrish:
This might also solve some of my concern with properties and relationships having different naming conventions. IANA mixes all kinds of naming conventions, but so far we only adopt "privacy-policy" from the relations. Perhaps properties can remain camelcased and we can hyphenate relationships for consistency?

This would mean only "pagelist" might potentially need to change to "page-list" (which would match EPUB but not DPUB-ARIA, though that hardly matters in this context). But leaving it as-is would fit just as well within the relations, too.

iherman:
Yes, I think separating those relationships to a different "thing" (table, list, or whatever) is a good idea. The quick reference may include property terms that come either from schema.org or from our own vocabulary, and it is unnecessary to refer to them as URL-s because they are only used as terms by the author. The relationship, alas!, is different.

(Unfortunately, I am not sure we can get around using URL-s for those at least for now...)

iherman:
(We should probably merge #425 before doing any change...)

mattgarrish:
Ya, this is a confusing one to piece apart. What we've done for some of these is define properties based on the relations, so we're not actually realizing that there is a property until the canonicalization step (although even that isn't clear for all).

In other words, what starts off as:

```
{
    "url": "toc.html",
    "rel": "contents"
}
```

after processing becomes:

```
{
   ...
   "contents": "html markup here",
   ...
}
```

since there are rules for compiling the table of contents, but they're not in canonicalization section.

But what about a resource that identifies itself as a privacy policy, does it become:

```
{
   ...
   "privacy-policy": "https://example.com/privacy.html",
   ...
}
```

If so, we're effectively minting new properties for every link relation that we compile up. But processing of privacy policy, cover and accessibility report is not detailed in the canonicalization steps, either, so when and where does it occur? Or are these supposed to be separate entities that a user agent might make itself aware of?

At any rate, this changes the dynamic of what I was proposing.

[...]

PR #428, opened 2019-04-19
separate resource relations from properties

mattgarrish:
This PR addresses #426 by making the following changes:

- The informative and structural "properties" are moved to a new section for resource relations and updated to remove any references to them being properties.

- A section for extending the relations list is included, but not sure if we need this or if it might be better in the LinkedResource definition.

- The web pub manifest requirements section now lists the recommended relations after the properties.

- Two new sections are added to the web pub manifest lifecycle: "extracting a toc" and "extracting a page list" - these were previously "extensions" of the properties, but all they define is the process for obtaining the markup so this seems like the most appropriate place to describe this.

- The "quick reference" section is removed and replaced by a properties index and a resource relations index. I'm not sure we need the latter, as the relations are all clearly found in the toc, but it's also harmless.

There's a fair bit of reorganizing and rewriting in this PR, so feedback welcome, especially if anyone notices any other areas in need of change I might not have noticed.

(I didn't change pagelist to page-list in this PR, as we can discuss whether hyphenated naming should be the standard for relations another time.)

Diff: https://tinyurl.com/y3wdtnfu
Preview: https://raw.githack.com/w3c/wpub/editorial/separate-relationships/index.html

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Apr 30, 2019, 3:44 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Frawcdn.githack.com%2Fw3c%2Fwpub%2F34c2ccddbbe4e8cae480f6945b0d8392b5e6be19%2Findex.html%3FisPreview%3Dtrue)

```
[33m📡 HTTP Error 520:[39m [36mhttps://rawcdn.githack.com/w3c/wpub/34c2ccddbbe4e8cae480f6945b0d8392b5e6be19/index.html[39m
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/wpub%23428.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4772,8 +4772,7 @@ dictionary LinkedResource {
 		<section id="properties-index" class="appendix informative">
 			<h4>Properties Index</h4>
 
-			<p>The following table lists where the properties defined in this specification are defined and
-				extended.</p>
+			<p>The following table identifies where the use of manifest properties is defined and extended.</p>
 
 			<table class="zebra">
 				<thead>
@@ -5076,7 +5075,7 @@ dictionary LinkedResource {
 		<section id="rel-index" class="appendix informative">
 			<h4>Resource Relations Index</h4>
 
-			<p>The following table lists where the resource relations are defined in this specification.</p>
+			<p>The following table identifies where the use of resource relations is defined.</p>
 
 			<table class="zebra">
 				<thead>
--- a/index.html
+++ b/index.html
@@ -2428,13 +2428,9 @@ partial dictionary PublicationManifest {
 					<h4>Extensions</h4>
 
 					<p>If additional relations beyond those defined in this specification need to be expressed, the <a
-							href="#dom-linkedresource-rel"><code>rel</code> property</a> can be extended in one of the
-						following ways:</p>
+							href="#dom-linkedresource-rel"><code>rel</code> property</a> can be extended through the use
+						of values defined in [[!iana-relations]].</p>
 
-					<ul>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -5052,6 +4769,366 @@ <h3>Cleaning up the data</h3>
...
+						<td></td>
+					</tr>
+				</tbody>
+			</table>
+		</section>
+		<section id="rel-index" class="appendix informative">
+			<h4>Resource Relations Index</h4>
+
+			<p>The following table lists where the resource relations are defined in this specification.</p>
+
```
iherman: This is just me being picky: we do _define_ these rel values _except_ for `contents`, which is a registered IANA term.

mattgarrish: described?

mattgarrish: In the interests of dodging the problem, I've changed the wording to make "the use of" what is defined - "The following table identifies where the use of resource relations is defined."

PR #432, opened 2019-05-01
change custom rel url values to tokens

### LINKED ISSUE #431: Change rel URLs to tokens
As noted in the [tag feedback](https://github.com/w3ctag/design-reviews/issues/344#issuecomment-485034489), the use of URLs for relations is an uncommon choice. It also means that if we don't standardize these with IANA before publishing, existing content will either have to be changed or user agents will forever be stuck supporting the URLs.

In the interests of clarity, can we just change:

- https://www.w3.org/ns/wp#cover -> cover
- https://www.w3.org/ns/wp#accessibility-report -> accessibility-report
- https://www.w3.org/ns/wp#pagelist -> pagelist (or page-list)

and deal with IANA registration separately?

mattgarrish:
This PR fixes #431 by changing all instances of cover, pagelist and accessibility-report from custom-namespaced URLs to plain tokens.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/432.html" title="Last updated on May 1, 2019, 5:38 PM UTC (eaed5bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/432/7a112da...eaed5bc.html" title="Last updated on May 1, 2019, 5:38 PM UTC (eaed5bc)">Diff</a>

### DIFF
--- a/experiments/separate_manifest/mobydick-simple.jsonld
+++ b/experiments/separate_manifest/mobydick-simple.jsonld
@@ -22,7 +22,7 @@
         "css/mobydick.css",
         {
             "type": "LinkedResource",
-            "rel": "https://www.w3.org/ns/wp#cover",
+            "rel": "cover",
             "url": "images/cover.jpg",
             "encodingFormat": "image/jpeg"
         },{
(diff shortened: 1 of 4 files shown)


mattgarrish:
I'm going to merge this, as it's a reasonably straight-forward change. Plus I noticed that we don't use a URL for the "publication" relation, so it also adds some consistency.

PR #436, opened 2019-05-04
Correct required "value" property

mattgarrish:
I noticed for localizable strings that we define "value" but require "@value". This PR assumes the former is correct.


mattgarrish:
Going ahead and merging this after double-checking the spec.

PR #439, opened 2019-05-09
Minor modification on the language and base direction section

iherman:
@dbaron had a comment in https://github.com/w3ctag/design-reviews/issues/344#issuecomment-485034489:

> The description of default language and base direction seems a bit confusing. It seems to imply it's a default for the manifest and the publication, but then there's a note that says it only applies to the manifest and not to the resources that comprise the publication.

I agree it was not clear. I have moved the relevant note to the beginning of the section to make the distinction between the language/direction of the publication and the language/direction of the individual resources. I hope this change is enough.

I have also added an explicit reference to the issue on whether lang/dir is inherited when the manifest is embedded.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/439.html" title="Last updated on May 9, 2019, 6:37 PM UTC (6d535a3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/439/91abe8d...6d535a3.html" title="Last updated on May 9, 2019, 6:37 PM UTC (6d535a3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1284,6 +1284,12 @@ partial dictionary PublicationManifest {
 							table of contents on the right hand side for publications whose natural base direction is
 							right-to-left).</p>
 
+						<p class="note">It is important to differentiate the language <em>of the publication</em>
+							from the language and the base direction of the individual resources that compose it. If
+							such resources are, for example, in HTML, the language and direction need to be set in
+							those resources, too. The language and base direction of the publication are not
+							inherited.</p>
+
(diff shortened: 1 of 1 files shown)


iherman:
@dbaron, are you o.k. with this change, w.r.t. your comment?

iherman:
@mattgarrish I would propose to merge this

PR #444, opened 2019-05-21
Algorithm and other cleanup

mattgarrish:
This PR initially started off to fix the address relics identified at the face-to-face, but expanded as I reviewed the spec to cover the following:

- removes the old paragraphs that described the type of expected resource at address (primary entry page being required now)
- clarifies the lifecycle algorithms due to primary entry page markup always being available because processing has to begin at it (i.e., shorthands for embedding are also available for linked manifests)
- adds an extension point for canonical manifest required fields as the address is specific to web publications
- makes it a terminating error if the address and the URL of the primary entry page don't match
- moves embedding/linking to Part 1 so that the generic lifecycle steps don't reference into web publications

There are also a number of smaller editorial fixes to the markup and wording.

Additionally, the warnings on the accessibility metadata values have been commented out, as this doesn't seem quite in keeping with schema.org use. It's perfectly valid to use other values than those listed in the preferred vocabulary.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/444.html" title="Last updated on May 21, 2019, 12:19 AM UTC (ee5752e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/444/91abe8d...ee5752e.html" title="Last updated on May 21, 2019, 12:19 AM UTC (ee5752e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -618,11 +618,11 @@ partial dictionary PublicationManifest {
 						<p>The base URL for relative URLs is determined as follows:</p>
 
 						<ul>
-							<li>In the case of an <a href="#wp-manifest-embedding">embedded manifest</a>, the base URL
-								is the <a data-cite="!html#the-base-element">document base URL</a>&#160;[[!html]] of the
+							<li>In the case of an <a href="#manifest-embed">embedded manifest</a>, the base URL is the
+									<a data-cite="!html#the-base-element">document base URL</a>&#160;[[!html]] of the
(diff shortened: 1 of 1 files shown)


mattgarrish:
One other significant change is that this PR clarifies that only the primary entry page can include a link to the manifest. Other resource can link to the PEP, but that seems to necessitate a new relation (and/or a rethink of whether "publication" is the relation for the manifest or the PEP).

PR #445, opened 2019-05-21
New i18n appendix

### LINKED ISSUE #443: Wording for D. Examples for bidirectional texts
D. Examples for bidirectional texts
https://w3c.github.io/wpub/#app-bidi-examples

The table in this appendix jumbles up a bunch of different things.

Do you want it to show how RLM/LRM can be prepended to a string in JSON to indicate the intended base direction? (If so, the examples should use javascript escapes, rather than HTML escapes.) (This is the kind of thing that rows 3 and 4 appear to be aiming at.)

Or do you want it to show how a consumer would wrap the string in markup/characters, to show how to apply the base direction in the target location when used? (which seems to be implied by the location of the link at the moment)

The second row of the table also shows how one can use characters to embed a different base direction *within* a string.  Do you really need this?  We don't mention string-internal bidi controls in the rest of the spec - only how to set the overall directional context for the string as a whole.

If you want to achieve more than one of the above, i think you will need to break up the table and add a bunch of introductory and explanatory text.

iherman:
Text provided by @r12a for #443.

Closes #443

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/445.html" title="Last updated on May 22, 2019, 5:00 AM UTC (6d8da4f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/445/f72a212...6d8da4f.html" title="Last updated on May 22, 2019, 5:00 AM UTC (6d8da4f)">Diff</a>


r12a:
I think it may be better to reword the 2nd sentence as:

"In cases where the first-strong heuristics would produce the wrong result, if the string is created with a prepended formatting character, the first-strong heuristics will produce the correct base direction for the string as a whole."

r12a:
Oh, please change back your 2nd edit.   I inserted the note between the two paragraphs intentionally, because the note applies only to what was the first paragraph, and NOT to the second.  

There are two scenarios described here: that of the producer (to which the note belongs) and that of the consumer (user agent), from which you also deleted what i think is an important sentence: 

"This could require wrapping additional markup or control characters around the string prior to display, in order to apply the base direction."

The wrapping of markup or control characters is NOT related to prepending an RLM/LRM.  It's done only when the string is inserted into a location where it will be displayed, and the control characters in question are almost certainly not RLM/LRM (but rather paired characters RLI, and friends).

Actually, i also realise now that the term 'control characters' is not correct either – it should be 'Unicode formatting characters'. (They aren't 'control' characters.)

iherman:
@r12a, I have made the reverse plus the few changes (moving the reference and change the control character term).

r12a:
Thanks. Did you see the other comment at https://github.com/w3c/wpub/pull/445#issuecomment-494447351 ?

Ok if you did – just checking.

iherman:
@r12a no, I missed https://github.com/w3c/wpub/pull/445#issuecomment-494447351 indeed.

However, I am not 100% sure which 'second sentence' you are referring to...

r12a:
In app D, introductory para.

iherman:
@r12 Ah, thanks, did it.

mattgarrish:
I made a couple of minor tweaks: the tables were missing some opening tags, and we shouldn't use the zebra css or tbody tags as these are layout tables (a11y).

PR #446, opened 2019-05-21
consolidate webidl

mattgarrish:
This PR consolidates all the webidl definitions in s. 2.3.2. It also removes the IDL index, which seems wholly redundant at this point.

@iherman is this the approach you were thinking, or something else?

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/446.html" title="Last updated on May 22, 2019, 2:59 PM UTC (d23529d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/446/620976b...d23529d.html" title="Last updated on May 22, 2019, 2:59 PM UTC (d23529d)">Diff</a>


mattgarrish:
We can also probably merge 2.3.1 and 2.3.2 if this is good, since there are no more pieces.

mattgarrish:
I've made some more changes, including removing the script.

Let me know if the first few paragraphs of 2.3 are clearer now that the webidl is not for exposing the data, which was also raised as a concern.

iherman:
Minor thing in 2.3:

> To simplify understanding of the manifest format

isn't it better to say

> To simplify the understanding of the manifest format

Otherwise I am fine with this text. 

There was a comment of Dave Baron to link 'back' from the definitions *into* the WebIDL. I am not sure respec does that automatically; I wonder whether it is possible to link at least into the relevant WebIDL dictionary from the respective text...

mattgarrish:
> to link 'back' from the definitions _into_ the WebIDL

What would these be, not repetitive links to "refer to the webidl definition", I hope? Links on the terms in the definition tables?

I took a look at appmanifest and don't see links back to the definition there.

mattgarrish:
I've added a script that sniffs out the dfns for publicationmanifest and links them back to the webidl. I assume that'll be sufficient?

iherman:
> What would these be, not repetitive links to "refer to the webidl definition", I hope? Links on the terms in the definition tables?
>
> I took a look at appmanifest and don't see links back to the definition there.

I do not know. The only reference is in https://github.com/w3ctag/design-reviews/issues/344#issuecomment-491056356.

But the script you have added seems to be perfect (pr preview seemed to crash, so I cannot see it...)

mattgarrish:
Ya, not sure what happened there. Maybe I did the two commits two quickly and crashed something. Anyway, seems to be back now.

iherman:
@mattgarrish I think that script of yours does the trick 100%. As far as I am concerned, it is a go...

PR #449, opened 2019-05-23
Change expected a11y property values to an informative note

mattgarrish:
This PR removes the "Expected Value" links for the accessibility properties from the definition table, since these were never meant as a restrictive set.

I've instead changed the paragraph after the definition table that directed people to the Web Schemas wiki to a note, and indicate in it that the expected values are listed there. The reference is now informative, as a result, and listed in the biblio file.

I've also fully removed the paragraphs about issuing warnings for unknown values from the manifest post-processing step. It's not invalid to use values other than those listed in the wiki; those are just the formally recognized values. Validating the values, and how best to inform about unknown values, shouldn't be in the user agent space.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/449.html" title="Last updated on May 23, 2019, 1:01 PM UTC (fc5ea8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/449/a5149c7...fc5ea8c.html" title="Last updated on May 23, 2019, 1:01 PM UTC (fc5ea8c)">Diff</a>

### DIFF
--- a/common/js/biblio.js
+++ b/common/js/biblio.js
@@ -128,5 +128,9 @@ var biblio = {
     "bibtex": {
          "title": "BibTeX Format Description",
          "href" : "http://www.bibtex.org/Format/"
+    },
+    "webschemas-a11y": {
+    	"title": "WebSchemas Accessibility",
+    	"href": "http://www.w3.org/wiki/WebSchemas/Accessibility"
     }
 }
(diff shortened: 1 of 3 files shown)


avneeshsingh:
This is appropriate. Good to have non-normative reference to schema.org wiki for values of accessibility metadata.

Issue #414, opened 2019-03-20
allow toc link markup to be preserved?


mattgarrish:
Closing those old toc issues jarred my memory that I don't think we ever took up the question about whether a text label has to be generated from the toc links or whether the child html content can be preserved.

Right now the algorithm references the accessible name calculation algorithm for retrieving a text label from the markup, but we may want to add the option to use the child markup itself?

Or it might be best to end the algorithm with the extraction of the child markup and leave it to the user agent to determine how best to present the label, with perhaps only guidance to use the accname algorithm when only a simple text label is desired?

iherman:
One of the reason we opted for the HTML based TOC is i18n. Ie, the chapter title may include ruby markup. I think that should be preserved, for example.

(I am not sure how this relates to the accessible name calculation, to be honest.)

mattgarrish:
Right, I think we went with text strings so that if the toc was also embedded directly in the manifest then the resulting structure would be the same regardless of how it was authored. Given embedding is not a consideration anymore, I think we should reconsider requiring text strings as the outcome.

dauwhe:
The ability of HTML to express more than text strings is important. Quite a bit of inline markup is semantically meaningful. Let's not give it up!

HadrienGardeur:
While we might "allow" it, I doubt that UAs will retain anything more than plain text. 

This has been discussed ad nauseam, but as @dauwhe has experimented himself, most UAs simply sanitize every string in a TOC.

mteixeira-wwn:
Here are a couple of examples of real EPUB 3.0.1 navigation markup. I mentioned in our call that we accept that these are nonstandard practices today, but I hope they're helpful insofar as they provide examples of eccentric, yet reasonably common, use cases in educational publishing.

> The examples below use certain attributes (`epub:type`, `data-`) that are used by our internal publishing automation tools or for cross-product interop. They can be ignored.

- A biochemistry book where units are organizational taxonomic labels. Chapter content documents still needed to be grouped by units. This is an example of `nav` entries without associated content document links.

```html
<li><span>INTRODUCTION</span>
    <ol>
        <li><a href="content/chapter01.xhtml" data-spine-index="10" data-nav-index="1" epub:type="chapter" data-page-start="2" data-page-end="5">Chapter 1: The Nature of Science</a>
            <ol>
[...]

PR #452, opened 2019-05-24
Allow retention of html tagging in toc extraction

mattgarrish:
This PR addresses #414 as follows:

- it provides the option to retain either the child content or generate a string (for the toc heading and entries)
- it only suggests using the accessible name computation algorithm to provide greater flexibility in how strings are generated

One downside of the flexibility is that it's not always clear when the child content will result in an empty string -- trimming whitespace isn't enough when retaining tagging. Trying to write out steps for that just gets us back to generating strings, so I've only stated in a parenthetical some things that would be involved, like removing whitespace and presentational tagging).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/pull/452.html" title="Last updated on May 27, 2019, 12:27 PM UTC (1204f60)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wpub/452/a5149c7...1204f60.html" title="Last updated on May 27, 2019, 12:27 PM UTC (1204f60)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4245,6 +4245,10 @@ dictionary LinkedResource {
 										processed if the value of the <em>toc</em>
 										<code>name</code> property is an empty string (i.e., no headings have yet been
 										encountered).</p>
+									<p>Whether a user agent sets the <code>name</code> to the descendant content of the
+										heading element, or generates a text string from it, depends on whether it will
+										re-use any descendant tagging in the presentation (e.g., to retain images,
+										MathML, ruby and other content that does not translate to text easily).</p>
 									<aside class="example" title="Visualization of the toc object with a heading">
 										<pre>
 {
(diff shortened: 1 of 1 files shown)


mattgarrish:
I added a description of why we list two options to the explanations for the relevant steps, but its a user agent processing section so every step is the user agent doing something. What clarification would you put at the top?

If you can get a real example of some useful markup to retain, that would be immeasurably better than anything I could fake.

iherman:
> 
> I added a description of why we list two options to the explanations for the relevant steps, but its a user agent processing section so every step is the user agent doing something. What clarification would you put at the top?
> 
> 

Ok, what you added may be indeed enough.

> If you can get a real example of some useful markup to retain, that would be immeasurably better than anything I could fake.
> 

I will

mattgarrish:
> Ok, what you added may be indeed enough.

If it helps, I've added this sentence to the second paragraph:

> In some steps, user agents are provided a choice in how to process the content to provide flexibility for different presentation models.

iherman:
@mattgarrish I would propose to merge this. Daihei or Naomi may give us an extra example using Ruby that we can add later, but we should avoid potential merge issues...

